PR #6, opened 2023-06-27
admin/fix-ci-and-upgrade-lint-dependencies

SeanLeRoy:
### Description of Changes

The CI was failing due to an issue with `isort` and `poetry`, upgrading `isort` to `>5.12.0` fixes this issue. Beyond that I noticed the versions were specified in both `.pre-commit-config.yaml` & `pyproject.toml`, it doesn't appear that `pre-commit` requires any of these dependencies to be installed prior to running `pre-commit` based on it ignoring whatever version I specify in `pyproject.toml` and solely using the version specified by the `rev` in the `pre-commit-config.yaml`. I tested this on MacOS as described below, but if I'm missing the value of `pyproject.toml` having the versions too let me know!

### Testing
Created a fresh virtual environment, installed dependencies, and ran `just lint` & `just test`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,14 +41,6 @@ Documentation = "https://bioio-devs.github.io/bioio"
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
 [project.optional-dependencies]
 lint = [
-  "black>=22.3.0",
-  "check-manifest>=0.48",
-  "flake8>=3.8.3",
-  "flake8-debugger>=3.2.1",
-  "flake8-pyprojecttoml",
-  "flake8-typing-imports>=1.9.0",
-  "isort>=5.7.0",
-  "mypy>=0.790",
   "pre-commit>=2.20.0",
 ]
 test = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -41,14 +41,6 @@ Documentation = "https://bioio-devs.github.io/bioio"
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
 [project.optional-dependencies]
 lint = [
-  "black>=22.3.0",
-  "check-manifest>=0.48",
```
SeanLeRoy, line 45: `check-manifest` wasn't being used in `pre-commit`, but instead it was installed without a specific version before `just install` was ran [here](https://github.com/bioio-devs/bioio/blob/main/.github/workflows/ci.yml#L32) so it seems safe to delete unless I am missing a usage.

BrianWhitneyAI (approved):

toloudis (approved):

pyproject.toml
evamaxfield: The only reason I include black is that VSCode will auto format on save when it is installed.

(I may be wrong about that, that is how I have set it up anyway)

Check manifest can go. It's very useful but used infrequently.

evamaxfield: Similarly, does VSCode not need any of these installed to have all the nice linting, formatting, type hinting, etc?

PR #7, opened 2023-06-29
feature/require minimum bioio-base for plugins

### LINKED ISSUE #2: Modify Plug-in Search Capabilities
### Feature Description

[plugins.py](https://github.com/bioio-devs/bioio/blob/main/bioio/plugins.py) is responsible for searching for plug-ins to use with this repository. Instead of letting any plug-in found be accepted, the search should reject any that don't meet the minimum `bioio-types` specified. 

### Use Case

Currently any plug-in matching a pattern criteria will be found and supplied which is great for now. However, as this repository evolves it is likely the contract between this repository and the readers (plug-ins) will too potentially leading to conflicts. We have an independent base that both this repository and the plug-ins depend on, [bioio-types](https://github.com/bioio-devs/bioio-types), so using bioio-types as a way to determine the compatibility should alleviate this.

SeanLeRoy:
### Related Issue 

This pull request resolves #2 where I requested plugins meet a minimum `bioio-types` version.

### Description of Changes

The search for plugins will now check the listed version specifications for this distribution (`bioio`) against each plugin ensuring that the listed minimum for `bioio` is <= the maximum listed for the plugin.

### Testing

No plugin has a minimum `bioio-types` yet, but I tested against other dependencies like `docutils` where it seems to perform as expected.


BrianWhitneyAI (approved):

bioio/plugins.py
```diff
@@ -53,24 +59,105 @@ def add_plugin(pluginentry: PluginEntry) -> None:
...
+        if (
+            min_compatible_bioio_base_version is not None
+            and max_bioio_base_version_for_plugin is not None
+            and semver.compare(
+                max_bioio_base_version_for_plugin,
+                min_compatible_bioio_base_version,
+            )
+            < 0
+        ):
+            print(
```
toloudis, line 140: at some point these prints should maybe be logs instead; not necessarily for this pr

toloudis (approved):
Looks good.  I wonder if it's unit testable with some mock data.

PR #9, opened 2023-07-10
Feature/add writers

### LINKED ISSUE #3: Add writers
### Feature Description

Add all the writers currently supported in `aicsimageio`

SeanLeRoy:
### Link to Relevant Issue

This pull request resolves #3 

### Description of Changes

This adds the writers from `aicsimageio`. The major difference between the writers and tests I copied from `aicsimageio` is that these no longer rely on `DefaultReader` or `OmeTiffReader` so I had to use `imageio` and `tifffile` more directly in the related writer and tests.

### DIFF
--- a/bioio/tests/conftest.py
+++ b/bioio/tests/conftest.py
@@ -16,19 +16,72 @@ Docs: https://docs.pytest.org/en/latest/example/simple.html
       https://docs.pytest.org/en/latest/plugins.html#requiring-loading-plugins-in-a-test-module-or-conftest-file
 """
 
+import logging
 import pathlib
+import shutil
 import typing
 
+import dask.array as da
+import numpy as np
 import pytest
+
+###############################################################################
+
+log = logging.getLogger(__name__)
+
+###############################################################################
 
 LOCAL_RESOURCES_DIR = pathlib.Path(__file__).parent / "resources"
 LOCAL_RESOURCES_WRITE_DIR = pathlib.Path(__file__).parent / "writer_products"
+
+
+def pytest_sessionstart(session: pytest.Session) -> None:
+    """
+    Called after the Session object has been created and
+    before performing collection and entering the run test suite loop.
+    """
+    if LOCAL_RESOURCES_WRITE_DIR.exists():
+        log.warning(
+            f"{LOCAL_RESOURCES_WRITE_DIR.absolute} should not exist at "
+            "start of tests, deleting now"
+        )
+        shutil.rmtree(LOCAL_RESOURCES_WRITE_DIR)
+
+
+def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
+    """
+    Called after whole test suite run finished, right before
+    returning the exit status to the system.
+    """
+    shutil.rmtree(LOCAL_RESOURCES_WRITE_DIR)
+
+
+@pytest.fixture
+def data_dir() -> pathlib.Path:
+    return pathlib.Path(__file__).parent / "data"
 
 
 def get_resource_full_path(filename: str) -> typing.Union[str, pathlib.Path]:
     return LOCAL_RESOURCES_DIR / filename
 
 
-@pytest.fixture
-def data_dir() -> pathlib.Path:
-    return pathlib.Path(__file__).parent / "data"
+def get_resource_write_full_path(filename: str) -> typing.Union[str, pathlib.Path]:
+    LOCAL_RESOURCES_WRITE_DIR.mkdir(parents=True, exist_ok=True)
+    return LOCAL_RESOURCES_WRITE_DIR / filename
+
+
+def np_random_from_shape(
+    shape: typing.Tuple[int, ...], **kwargs: typing.Any
+) -> np.ndarray:
+    return np.random.randint(255, size=shape, **kwargs)
+
+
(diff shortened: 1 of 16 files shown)


pyproject.toml
```diff
@@ -29,6 +29,14 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
   "bioio-base>=0.1.1",
+  "dask[array]>=2021.4.1",
```
SeanLeRoy, line 32: Lots of dependencies added here that are specific to the `writers`. Does it seem worth it to move these to an extra so they are installable like `bioio[writers]` instead of by default or is that just added complexity for not much gain?

bioio/tests/writers/test_timeseries_writer.py
```diff
@@ -0,0 +1,174 @@
...
+
+from ..conftest import array_constructor, get_resource_write_full_path
+
+
+@array_constructor
+@pytest.mark.parametrize(
+    "write_shape, write_dim_order, read_shape",
+    [
+        # TODO: Failing currently, needs work,
+        # see https://github.com/bioio-devs/bioio/issues/10
```
SeanLeRoy, line 22: These two commented out tests pass on `aicsimageio` (technically) see the issue for why this was commented out.

bioio/tests/conftest.py
```diff
@@ -16,19 +16,72 @@
...
+
+###############################################################################
+
 LOCAL_RESOURCES_DIR = pathlib.Path(__file__).parent / "resources"
 LOCAL_RESOURCES_WRITE_DIR = pathlib.Path(__file__).parent / "writer_products"
 
 
-def get_resource_full_path(filename: str) -> typing.Union[str, pathlib.Path]:
-    return LOCAL_RESOURCES_DIR / filename
+def pytest_sessionstart(session: pytest.Session) -> None:
```
SeanLeRoy, line 38: Actually rethinking this I think I'll change this to use tempdir. Maybe we can write out an `example.txt` to tempdir too and avoid having any files download for this package thus far.

pyproject.toml
toloudis: dask is an interesting one - I can see many projects importing dask on their own too.  Personally I don't have a strong opinion other than "keep it simple" - start with this and we can change things later if we need to separate it into an extra.

pyproject.toml
```diff
@@ -29,6 +29,14 @@ classifiers = [
...
+  "fsspec>=2022.8.0",
+  "imageio[ffmpeg]>=2.11.0,<2.28.0",
+  "numpy>=1.16.0,<2.0.0",
+  "ome-types>=0.3.3",
```
toloudis, line 36: I think ome-types 0.3.4, but also see this https://forum.image.sc/t/ome-types-rewritten-requesting-alpha-testers/83174

toloudis: https://ome-types.readthedocs.io/en/latest/migration/

toloudis (approved):

pyproject.toml
tlambert03: while that migration guide looks like a lot, you should also know that I routinely [test against aicsimageio](https://github.com/tlambert03/ome-types/actions/runs/5545243407/jobs/10123881180) ... and you should be able to use v0.4.0 without any errors (albeit plenty of name change warnings which that migration guide would help with).

I don't _think_ it should cause any ecosystem difficulties if you pin to >0.4.0 ... since v0.4.0 should work for any libraries that haven't updated

toloudis: I basically agree, we might as well go with the latest if there are no obvious problems 🚀

tlambert03: happy to help with that.  do you want me to open a PR to remove all the warnings?

tlambert03: btw... i'm also working on pydantic2 support (https://github.com/tlambert03/ome-types/pull/205) ... and it's speedy!

Screen Shot 2023-07-13 at 5 50 16 PM

(though, I don't think you should pin pydantic yet)

toloudis: @SeanLeRoy should chime in, but I'll just say that one other option is to defer this dependency version update till we move further along on getting bioio ready.   But I'm comfy putting the change both here and aicsimageio anytime

tlambert03: yep, sounds good either way.  just ping me if I can be of help

SeanLeRoy: I'm willing to try out migrating to `ome-types>=0.4.0`. I'll give it a shot as part of this and update here with my findings though it seems like it should be an easy migration

SeanLeRoy: Seemed easy enough 👍 , looks like as part of this `lxml` can be dropped as a direct dependency of this package and installed as an extra from `ome-types` which I did in [#1dcd7ae](https://github.com/bioio-devs/bioio/pull/9/commits/7ac3d9e2939fdcda78a5e18e8cc1d0ad7cdca76f)

toloudis (approved):
I wish there were an explicit declaration of `tmp_path` somewhere.  As a newbie it looks confusing because where did it come from?   But I am not complaining, it's incredibly convenient.

BrianWhitneyAI (approved):
Looks good to me! need to add the changes to ome_zarr_writer once they get merged but that is for a later date.

PR #13, opened 2023-08-02
Migrate python versions

SeanLeRoy:
### Description of Changes

Add CI support for python 3.11 and removes support for python 3.8


tlambert03:
Are you dropping 3.8 because of [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html), or because there are actually things you want that are only in 3.9+ and it would actively behoove you to drop 3.8?

pyproject.toml
```diff
@@ -22,13 +22,13 @@ authors = [
...
   "Natural Language :: English",
-  "Programming Language :: Python :: 3.8",
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
 ]
 dynamic = ["version"]
 dependencies = [
-  "bioio-base>=0.1.1",
+  "bioio-base>=0.3.0",
```
toloudis, line 31: bioio-base is already versioned? 😄

toloudis (approved):

pyproject.toml
SeanLeRoy: Yep! 🎉

PR #19, opened 2023-11-02
use correct package name for bioio-base

toloudis:
I think we can't really use bioio without this change. This is required for get_plugins to work, and get_plugins is currently reqiured in order to have any file format support.

### DIFF
--- a/bioio/plugins.py
+++ b/bioio/plugins.py
@@ -22,7 +22,7 @@ from bioio_base.reader_metadata import ReaderMetadata
 ###############################################################################
 
 BIOIO_DIST_NAME = "bioio"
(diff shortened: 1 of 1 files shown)


bioio/plugins.py
```diff
@@ -22,7 +22,7 @@
 ###############################################################################
 
 BIOIO_DIST_NAME = "bioio"
-BIOIO_BASE_DIST_NAME = "bioio-types"  # TODO: Rename to bioio-base
+BIOIO_BASE_DIST_NAME = "bioio-base"  # TODO: Rename to bioio-base
```
evamaxfield, line 25: Can remove the TODO

toloudis: oops

evamaxfield (approved):
Thank you!

SeanLeRoy (approved):

Issue #18, opened 2023-11-02
Features / Bugs / Eva Notes


evamaxfield:
Wanted to give the new libraries a whirl and figured I would make a couple of notes as I go along. I will try to resolve these issues / features this weekend or something as many of them are "optional" / maybe nice for dev and debugging experience.

---

1. `BIOIO_BASE_DIST_NAME` needs to be updated from `bioio-types` to `bioio-base`

2. It would be nice to have access to what plugin read the file somehow. Currently on image read, the `__repr__` returns: `<BioImage [Reader: Reader, Image-is-in-Memory: False]>` but it might be better to return `<BioImage [plugin: {plugin_name}, Image-is-in-Memory: False]>`, esp. because most plugins will generally all use the same class name of `Reader` instead of say `OmeTiffReader`, better might be `<BioImage [reader_path: {python_module_path_of_reader}, Image-is-in-Memory: False]>`

3. `plugin_cache` should be made into a set, dict, or deduped list:

```python
In [2]: from bioio.plugins import get_plugins

In [3]: get_plugins()
Out[3]: 
[PluginEntry(entrypoint=EntryPoint(name='bioio-tifffile', value='bioio_tifffile', group='bioio.readers'), metadata=<class 'bioio_tifffile.reader_metadata.ReaderMetadata'>, timestamp=1698949015.0319347),
 PluginEntry(entrypoint=EntryPoint(name='bioio-ome-tiff', value='bioio_ome_tiff', group='bioio.readers'), metadata=<class 'bioio_ome_tiff.reader_metadata.ReaderMetadata'>, timestamp=1698949015.0319347),
 PluginEntry(entrypoint=EntryPoint(name='bioio-ome-zarr', value='bioio_ome_zarr', group='bioio.readers'), metadata=<class 'bioio_ome_zarr.reader_metadata.ReaderMetadata'>, timestamp=1698949015.0319347),
 PluginEntry(entrypoint=EntryPoint(name='bioio-lif', value='bioio_lif', group='bioio.readers'), metadata=<class 'bioio_lif.reader_metadata.ReaderMetadata'>, timestamp=1698949015.0319347)]

In [4]: get_plugins()
Out[4]: 
[PluginEntry(entrypoint=EntryPoint(name='bioio-tifffile', value='bioio_tifffile', group='bioio.readers'), metadata=<class 'bioio_tifffile.reader_metadata.ReaderMetadata'>, timestamp=1698949015.0319347),
[...]

PR #20, opened 2023-11-02
improve error string

toloudis:
error string was previously showing bioio_base.exceptions.UnsupportedFileFormatError: BioImage does not support the image: '<class 'str'>'.

This code change will utilize the image class type only if the image is not a file path and otherwise print the path.


evamaxfield (approved):
Resolves item 2 from #18!

Working fast :heart:

SeanLeRoy (approved):

PR #22, opened 2023-11-03
Auto retrieve plugins + adjust caching

SeanLeRoy:
This change makes `BioImage` automatically retrieve plugin each time it determines which plugin/reader to use.

I made caching an opt-in feature since I am a bit worried about users being confused by cached plugins.

### DIFF
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -13,7 +13,7 @@ import xarray as xr
 from ome_types import OME
 
 from .ome_utils import generate_ome_channel_id
-from .plugins import get_plugins, PluginEntry
+from .plugins import PluginEntry, get_plugins
 
 ###############################################################################
 
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -12,7 +12,7 @@ import xarray as xr
 from ome_types import OME
 
 from .ome_utils import generate_ome_channel_id
-from .plugins import plugins_by_ext
+from .plugins import get_plugins
 
 ###############################################################################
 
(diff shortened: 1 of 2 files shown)


toloudis:
I wonder if you could just call get_plugins once in __init__.py and just keep that as the cache.  Then just make sure that any additional calls blow away what was there before and completely replace the list.

SeanLeRoy:
> I wonder if you could just call get_plugins once in `__init__.py` and just keep that as the cache. Then just make sure that any additional calls blow away what was there before and completely replace the list.

I think that is the same as what would happen after this changeset. `determine_plugin` is only called once in the `__init__` and the only other time it would be called is if a user explicitly called it as a static method of `BioImage` during which time they could directly supply the cache arg (though I imagine if you are calling it directly it is for some sort of debug purpose in which you would want the cache to be reset each time).

toloudis:
> > I wonder if you could just call get_plugins once in `__init__.py` and just keep that as the cache. Then just make sure that any additional calls blow away what was there before and completely replace the list.
> 
> I think that is the same as what would happen after this changeset. `determine_plugin` is only called once in the `__init__` and the only other time it would be called is if a user explicitly called it as a static method of `BioImage` during which time they could directly supply the cache arg (though I imagine if you are calling it directly it is for some sort of debug purpose in which you would want the cache to be reset each time).

I was actually referring to calling it globally in the bioio package init. (which I guess is invoked the first time you import anything from bioio? or every time?) But I think what you said makes sense, it's kinda the same result.

toloudis (approved):

evamaxfield (approved):
Thank you!!

PR #42, opened 2024-04-29
relax the enforce_exists check because it will be done later

### LINKED ISSUE #41: HTTP URL for ZARR fails to be validated by `determine_plugin` for bioio-ome-zarr
Description: see title

# Code to reproduce
```python
from bioio import BioImage
path = "https://allencell.s3.amazonaws.com/aics/nuc_morph_data/data_for_analysis/baseline_colonies/20200323_09_small/raw.ome.zarr"
image = BioImage(path)  # FileNotFoundError
print(image.get_image_dask_data())
```

## Workaround
If the OME ZARR reader is specified explicitly, `determine_plugin` is bypassed and the image can be loaded. `bioio.BioImage(path, reader=bioio_ome_zarr.Reader)`.

# Environment
I'm testing against bioio-ome-zarr [PR 17](https://github.com/bioio-devs/bioio-ome-zarr/pull/17) which improves support for reading from S3.

toloudis:
### Link to Relevant Issue

This pull request resolves #41 

### Description of Changes

Related to #40 .
companion PR to bioio-ome-zarr: https://github.com/bioio-devs/bioio-ome-zarr/pull/20
Don't enforce_exists until specific Readers are trying to check for file support.

A side effect of this is that some bad file paths will raise a different kind of exception.  (see modified unit test)

BONUS FIX:  macos-latest CI build was failing due to M1 support for imageio-ffmpeg.


evamaxfield (approved):
In general I approve. Still wish we could give back a specific "this file wasn't found" vs "this file isn't supported" error but I understand

BrianWhitneyAI (approved):

pgarrison (approved):

toloudis:
> In general I approve. Still wish we could give back a specific "this file wasn't found" vs "this file isn't supported" error but I understand

There is a way to do this - it's possible for the `ReaderClass.is_supported_image` to raise it now, so I added a couple lines to catch and rethrow it.
No unit test gets this because there are no plugin readers installed when we run the unit tests.
But I verified by adding a tiff plugin and sending a fake .tiff file path and it did raise FileNotFound

Issue #39, opened 2024-04-09
Add ArrayLikeReader


toloudis:
### Feature Description

BioImage should be able to wrap an already in memory numpy array.

### Use Case

This previously existed in aicsimageio and some scientists have asked for it.

### Solution

Bring this implementation over:
https://github.com/AllenCellModeling/aicsimageio/blob/main/aicsimageio/readers/array_like_reader.py
Add it to bioio.
Plugin selection can check for arraylike first and apply this reader if a numpy array is passed in instead of a path/url etc.

BrianWhitneyAI:
Some clarifying questions:

1) This would not be creating a new reader plugin this would be adding the arraylike reader to the bioio repository 
2) The check would most likely occur [here](https://github.com/bioio-devs/bioio/blob/3b3dc54789263295b53a7df2f71a0a54c038a0e4/bioio/bio_image.py#L151)

PR #47, opened 2024-07-02
feature/array-like-reader

BrianWhitneyAI:
## Description 
The purpose of this PR is to port over the ArrayLikeReader from aicsimageio, resolving #39. In general, it seems a better idea to have the arraylike reader as a builtin to bioio rather than a separate plugin.


bioio/bio_image.py
```diff
@@ -125,6 +129,22 @@ class BioImage(biob.image_container.ImageContainer):
     by either instantiating the reader independently or using the `.reader` property.
     """
 
+    @staticmethod
+    def get_array_like_plugin() -> PluginEntry:
+        """
+        Create and return a PluginEntry for ArrayLikeReader.
+        """
+        entrypoint = EntryPoint(
```
BrianWhitneyAI, line 137: I am not sure how important this information is. These are more placeholders right now.

bioio/tests/test_array_like_reader.py
```diff
@@ -1139,3 +1139,9 @@ def test_aicsimage_from_array(
             dict,
         ),
     )
+
+
```
BrianWhitneyAI, line 1143: @toloudis is this sort of what you had in mind for testing the type offset?

bioio/bio_image.py
```diff
@@ -125,6 +129,22 @@ class BioImage(biob.image_container.ImageContainer):
     by either instantiating the reader independently or using the `.reader` property.
     """
 
+    @staticmethod
+    def get_array_like_plugin() -> PluginEntry:
```
toloudis, line 133: I suggest moving this whole function to plugins.py

SeanLeRoy (approved):
Awesome job 🎉

bioio/bio_image.py
BrianWhitneyAI: Resolved in [0c6bb8e](https://github.com/bioio-devs/bioio/pull/47/commits/0c6bb8ecb19ceef7b1e17b50d9599ace6432b9dc)

bioio/tests/test_array_like_reader.py
toloudis: Yep!  passing file path into arraylikereader raises exception.  LGTM!

toloudis (approved):
Thanks for bringing this over from aicsimageio!

bioio/array_like_reader.py
```diff
@@ -0,0 +1,470 @@
...
+        if isinstance(image, list):
+            return all(
+                isinstance(scene, (np.ndarray, da.Array, xr.DataArray))
+                for scene in image
+            )
+
+        return isinstance(image, (np.ndarray, da.Array, xr.DataArray))
+
+    @staticmethod
+    def _guess_dim_order(shape: Tuple[int, ...]) -> str:
```
evamaxfield, line 122: Just curious. I think this may be the third time we have this code in multiple places? (Base) TiffReader, this one, and somewhere else?? Should we centralize this to a standard function that is called `guess_dims_from_shape` that all of these different Readers call?

evamaxfield (approved):
Nice work! Thanks for adding this back in.

Only one comment that is an open discussion for everyone on if we should create a standard guess dims function rather than copy pasting it around to different readers.

PR #48, opened 2024-07-03
Feature/ome zarr writer v2

### LINKED ISSUE #46: Add new ome-zarr-writer
### Feature Description

We have a new ome-zarr-writer implementation that works much better for converting our large time series data.  We'd like to make that a part of bioio now. 

### Use Case

We have lots of data to convert to ome-zarr.  The current ome zarr writer is unstable and doesn't work well on large data.

### Solution

Clean up the code from here:
https://github.com/allen-cell-animated/cellbrowser-tools/blob/nucmorph/cellbrowser_tools/ome_zarr_writer.py
and add it to this repo.   
Add documentation as needed.    

Question:  How should we deprecate the old zarr writer that already exists in here?

toloudis:
### Link to Relevant Issue

Resolves #46

We have been developing a new ome-zarr-writer that works better for our long time series data.  We want to bring that code into this repo and make it a part of bioio.
 
### Description of Changes

Add new ome zarr writer.  Do some code cleanup, and make sure it is well commented and documented.
Add unit tests for it.

**This PR is intended to be like an "initial add", bringing the new code in without any functional changes, apart from code formatting and commenting, so that it can be a direct replacement in our ome-zarr-conversion codebase.**


.pre-commit-config.yaml
```diff
@@ -1,31 +1,30 @@
...
-    rev: 23.3.0
-    hooks:
-      - id: black
+    - repo: https://github.com/PyCQA/flake8
+      rev: 6.1.0
+      hooks:
+          - id: flake8
+            additional_dependencies:
+                - flake8-typing-imports>=1.9.0
+                - flake8-pyprojecttoml
```
evamaxfield, line 25: Just as a general note, this is all fine but [ruff](https://github.com/astral-sh/ruff) does all of these in one except for `black`. My pre-commit configs are now: black, ruff, mypy

bioio/writers/__init__.py
```diff
@@ -2,9 +2,11 @@
...
 from .ome_tiff_writer import OmeTiffWriter
-from .ome_zarr_writer import OmeZarrWriter
+from .ome_zarr_writer import OmeZarrWriter as OmeZarrWriter_old
+from .ome_zarr_writer_2 import OmeZarrWriter
 
 __all__ = [
     "OmeTiffWriter",
     "OmeZarrWriter",
+    "OmeZarrWriter_old",
 ]
```
evamaxfield, line 12: Is our old reader better for large Z stack data while this one is better for timeseries or is this one just better overall?

.pre-commit-config.yaml
toloudis: The point of this PR is not to change any build or automation infrastructure.   But if you think we should use ruff then let's make a new ticket for that.

bioio/writers/__init__.py
toloudis: Good question.  (Note this is still a draft PR and I am not expecting that this `_old` suffix is really best.  It's less of a breakage to call the new one OmeZarrWriter2 and leave the old one with same name.)

Here are some of the key differences:
* The new one is built on tooling for computing multiscales that is more stable than what the old one uses
* The new one works better for bigger data because it does things in smaller steps. There is no one one-size-fits-all "write this data array to this url" function in the new writer -- at least not yet.   
* The new code is coming from another repo where it's been used to do some bulk data conversions internally already.  
* The new one is currently providing functions that expect to deal with time series of zstacks where the zstacks still might reasonably fit in memory.  
* The new one is currently assuming 5D data coming from bioio.BioImage, or sequences of images (one per T), or (in progress) a 5D ArrayLike.

.pre-commit-config.yaml
evamaxfield: Created a new issue: https://github.com/bioio-devs/bioio/issues/50

toloudis: I have been hearing good things.. 🦀

pyproject.toml
```diff
@@ -32,12 +32,14 @@ dependencies = [
...
   "imageio[ffmpeg]>=2.11.0,<2.28.0",
+  "ngff-zarr>=0.8.2",
   "numpy>=1.21.0,<2.0.0",
   "ome-types[lxml]>=0.4.0",
   "ome-zarr>=0.6.1",
   "semver>=3.0.1",
   "tifffile>=2021.8.30",
-  "zarr>=2.6.0,<3.0.0",
+  # this pin <2.18.0 should be temporary and is maybe related to https://github.com/zarr-developers/zarr-python/issues/1891
+  "zarr>=2.6.0,<2.18.0",
```
toloudis, line 42: sadly, without this version pin, unit tests were failing.  I don't know what the fix is but we should probably have a "reminder" issue to try to unpin it again later.

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+            # Generate a metadata dict and write it to the zarr.
+            meta = writer.generate_metadata(
+                image_name="my_image_name",
+                channel_names=my_channel_names,
+                physical_dims=physical_scale,
+                physical_units=physical_units,
+                channel_colors=my_channel_colors,
+            )
+            writer.write_metadata(meta)
+    """
```
toloudis, line 361: the above example is the typical usage/code flow of how to use this writer

.github/workflows/ci.yml
```diff
@@ -30,83 +30,83 @@ jobs:
...
+
   # Check tests pass on multiple Python and OS combinations
   test:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.9, "3.10", "3.11"]
-        os: [ubuntu-latest, macOS-latest, windows-latest]
+        python-version: [3.9, "3.10", "3.11", "3.12"]
```
BrianWhitneyAI, line 40: nit. I believe the quotations are only necessary for versions with a trailing zero?

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+        raise ValueError("shape must be a 5-tuple in TCZYX order")
+
+    itemsize = np.dtype(dtype).itemsize
+    # let's start by just mandating that chunks have to be no more than
+    # 1 T and 1 C
+    chunk_size = (1, 1, shape[2], shape[3], shape[4])
+    while prod(chunk_size) * itemsize > memory_target:
+        # chop every dim in half until they get down to 1
+        chunk_size = (
+            max(chunk_size[0] // 2, 1),
```
BrianWhitneyAI, line 52: These are guaranteed to be 1 correct?

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+
+    Returns
+    -------
+    Resized image.
+    """
+    factors = np.array(output_shape) / np.array(image.shape).astype(float)
+    # Rechunk the input blocks so that the factors achieve an output
+    # blocks size of full numbers.
+    better_chunksize = tuple(
+        np.maximum(1, np.round(np.array(image.chunksize) * factors) / factors).astype(
```
BrianWhitneyAI, line 95: nit. Different methods of finding maximum throughout file

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+    Chunk size tuple.
+    """
+    if len(shape) != 5:
+        raise ValueError("shape must be a 5-tuple in TCZYX order")
+
+    itemsize = np.dtype(dtype).itemsize
+    # let's start by just mandating that chunks have to be no more than
+    # 1 T and 1 C
+    chunk_size = (1, 1, shape[2], shape[3], shape[4])
+    while prod(chunk_size) * itemsize > memory_target:
```
BrianWhitneyAI, line 49: I think this could be simplified to 
```   
while prod(chunk_size) * itemsize > memory_target:
    chunk_size = tuple(max(cs // 2, 1) for cs in chunk_size)
```

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+
+    output_slices = tuple(slice(0, d) for d in output_shape)
+    output = da.map_blocks(
+        resize_block, image_prepared, dtype=image.dtype, chunks=block_output_shape
+    )[output_slices]
+    return output.rechunk(image.chunksize).astype(image.dtype)
+
+
+def _pop_metadata_optionals(metadata_dict: dict) -> dict:
+    for ax in metadata_dict["axes"]:
```
BrianWhitneyAI, line 130: maybe
`metadata_dict["axes"] = [ax for ax in metadata_dict["axes"] if ax.get("unit") is not None]`

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+    nlevels:
+        Number of levels to return. The first level is the
+        original lvl0shape.
+
+    Returns
+    -------
+    List of shapes of all nlevels.
+    """
+    shapes = [lvl0shape]
+    for i in range(nlevels - 1):
```
BrianWhitneyAI, line 250: Maybe 
```
for _ in range(nlevels - 1):
  shapes.append(tuple(max(int(s / sc), 1) for s, sc in zip(shapes[-1], scaling)))
```

BrianWhitneyAI (approved):
This is awesome! Nice work!

bioio/writers/ome_zarr_writer_2.py
evamaxfield: Does this also resolve the desire to be able to append new data to the zarr store over time? I.e. while a microscope is taking a timelapse, can we store each timepoint as they come in?

If its not possible, no worries, if it is possible, I would say that is worth highlighting

evamaxfield (approved):
Generally all looks good to me! Nice work. My biggest nitpick is to store this in a different location rather than calling it "ome_zarr_writer_2" maybe `bioio.experimental.writers.ome_zarr_writer`?

bioio/writers/ome_zarr_writer_2.py
toloudis: This code isn't made for modifying an existing one, I believe it's only been tested for when you are ready to write everything all at once... the use case of converting pre-existing files in other formats.  But it might work in the scenario you describe!

.github/workflows/ci.yml
toloudis: I can remove em - I was following the pre-existing pattern which already had 3.11 in quotes.

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+                    "min": float(channel_minmax[i][0]),
+                    "start": float(channel_minmax[i][0]),
+                },
+            }
+        )
+
+    omero = {
+        "id": 1,  # ID in OMERO
+        "name": image_name,  # Name as shown in the UI
+        "version": "0.4",  # Current version
```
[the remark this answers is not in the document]
toloudis: This is intended to be the ome-ngff spec version number.  I agree I can hoist it up to a symbolic constant but it is also potentially intimately connected to the names in this exact dict.

toloudis:
> Generally all looks good to me! Nice work. My biggest nitpick is to store this in a different location rather than calling it "ome_zarr_writer_2" maybe `bioio.experimental.writers.ome_zarr_writer`?

Actually I would suggest that this writer code is LESS experimental than the other, since it's been used (from a different repo) to do quite a lot of bulk conversions already.

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -0,0 +1,688 @@
...
+) -> dict:
+    if len(dims) != 5:
+        raise ValueError("dims must be a 5-tuple in TCZYX order")
+    return {"t": dims[0], "c": dims[1], "z": dims[2], "y": dims[3], "x": dims[4]}
+
+
+def resize(
+    image: da.Array, output_shape: Tuple[int, ...], *args: Any, **kwargs: Any
+) -> da.Array:
+    r"""
```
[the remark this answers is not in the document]
toloudis: I can remove it

bioio/writers/ome_zarr_writer_2.py
toloudis: I'm not going to modify this one as it is very specific for the downsampling code and is borrowed from a trusted implementation

bioio/writers/ome_zarr_writer_2.py
toloudis: nope, mypy hates that due to the enforcement of 5D tuples here.

bioio/writers/ome_zarr_writer_2.py
toloudis: if I did that it would change the behavior of this code.  your suggested change is to only include every axis that has a unit.  the point of this code is to go through each "axis" and remove the "unit" key if it happened to have a None value.  This function is called "pop metadata optionals" - it's just trying to delete keys that aren't holding meaningful data and the spec says are optional.

PR #55, opened 2024-07-18
bugfix/longest-suffix-wins

### LINKED ISSUE #54: bioio-tifffile takes precedence over bioio-ome-tiff for ome-tiffs
### Describe the Bug

[bioio-tifffile](https://github.com/bioio-devs/bioio-tifffile) takes precedence over [bioio-ome-tiff](https://https://github.com/bioio-devs/bioio-ome-tiff) even for `ome.tiff` files.

### Expected Behavior

OME-TIFFs should be correctly picked up by `bioio-ome-tiff`

### Reproduction

I am in `napari_bioio` but its the same test data:

```python
from bioio import BioImage

img = BioImage("napari_bioio/tests/resources/variance-cfe.ome.tiff")
img
```

`img` returns as `<BioImage [plugin: bioio-tifffile installed at 2024-07-18 16:09:20.787038, Image-is-in-Memory: False]>`

### Environment

-   OS Version: _[e.g. macOS 11.3.1]_
-   bioio Version: _[e.g. 0.5.0]_

```
❯ pip freeze | grep bioio

bioio==1.0.3
bioio-base==1.0.4
bioio-imageio==1.0.0
bioio-ome-tiff==1.0.0
bioio-ome-zarr==1.0.1
bioio-tiff-glob==1.0.0
bioio-tifffile==1.0.0
```

Already determined that the issue is that plugin list isn't sorted into longest matching suffix first (i.e. longer suffixes are more specific and we should try them first in the case of plugins handling multiple filetypes)

evamaxfield:
…ost specific plugins are matched first

### Link to Relevant Issue

This pull request resolves #54 

### Description of Changes

Sort the list of plugins by suffix length prior to determining the plugin reader to ensure we find the most specific viable readers first

### DIFF
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -164,9 +164,6 @@ class BioImage(biob.image_container.ImageContainer):
 
             # Check for extension in plugins_by_ext
             for format_ext, plugins in plugins_by_ext.items():
-                # Remove "." prefix if already there
-                if format_ext.startswith("."):
-                    format_ext = format_ext[1:]
                 if path.lower().endswith(f".{format_ext}"):
                     for plugin in plugins:
                         ReaderClass = plugin.metadata.get_reader()
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -164,7 +164,7 @@ class BioImage(biob.image_container.ImageContainer):
 
             # Check for extension in plugins_by_ext
             for format_ext, plugins in plugins_by_ext.items():
-                if path.lower().endswith(f".{format_ext}"):
+                if path.lower().endswith(format_ext):
                     for plugin in plugins:
                         ReaderClass = plugin.metadata.get_reader()
                         try:
--- a/bioio/plugins.py
+++ b/bioio/plugins.py
@@ -17,7 +17,7 @@ else:
     from importlib.metadata import entry_points, EntryPoint, requires
 
 import time
-from typing import Dict, List, NamedTuple, Optional, Tuple
+from typing import Dict, List, NamedTuple, Optional, Tuple, Union
 
 from bioio_base.reader import Reader
 from bioio_base.reader_metadata import ReaderMetadata
@@ -38,9 +38,9 @@ class PluginEntry(NamedTuple):
 
 
 # global cache of plugins
-plugins_by_ext_cache: (
-    Dict[str, List[PluginEntry]] | OrderedDict[str, List[PluginEntry]]
-) = {}
+plugins_by_ext_cache: Union[
+    Dict[str, List[PluginEntry]], OrderedDict[str, List[PluginEntry]]
+] = {}
 
 
 def check_type(image: ImageLike, reader_class: Reader) -> bool:
--- a/bioio/plugins.py
+++ b/bioio/plugins.py
@@ -7,7 +7,6 @@ import sys
 from collections import OrderedDict
 from datetime import datetime
 from pathlib import Path
-from pprint import pprint
 from typing import get_args
 
 import semver
@@ -231,7 +230,6 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
                 # And worse can cause this plugin list to have multiple entries
                 # for the same extension (e.g. ".tif" and "tif")
                 if ext.startswith("."):
-                    print(f"Plugin: {plugin_entry} includes leading '.' in ext: {ext}")
                     ext = ext[1:]
 
                 # Start a new list of plugins for ext if it doesn't exist
@@ -242,12 +240,6 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
                 # insert in sorted order (sorted by most recently installed)
                 pluginlist = plugins_by_ext[ext]
                 insert_sorted_by_timestamp(pluginlist, plugin_entry)
-
-    print("Before sorting:")
-    pprint(plugins_by_ext)
-
-    print()
-    print()
 
     # Dictionary values (the lists of plugin entries) have already been sorted
     # by timestamp due to the "insert_sorted_by_timestamp" function
@@ -264,9 +256,6 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
             reverse=True,
         )
     )
-
-    print("After sorting:")
-    pprint(plugins_by_ext)
 
     # Save copy of plugins to cache then return
     plugins_by_ext_cache.clear()
--- a/bioio/plugins.py
+++ b/bioio/plugins.py
@@ -233,6 +233,11 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
     # Save copy of plugins to cache then return
     plugins_by_ext_cache.clear()
     plugins_by_ext_cache.update(plugins_by_ext)
+
+    # Sort dict from longest to shortest key
+    plugins_by_ext = dict(
+        sorted(plugins_by_ext.items(), key=lambda x: len(x[0]), reverse=True)
+    )
     return plugins_by_ext
 
 
--- a/bioio/plugins.py
+++ b/bioio/plugins.py
@@ -4,8 +4,10 @@
 import os
 import re
 import sys
+from collections import OrderedDict
 from datetime import datetime
 from pathlib import Path
+from pprint import pprint
 from typing import get_args
 
 import semver
@@ -37,7 +39,9 @@ class PluginEntry(NamedTuple):
 
 
 # global cache of plugins
-plugins_by_ext_cache: Dict[str, List[PluginEntry]] = {}
+plugins_by_ext_cache: (
+    Dict[str, List[PluginEntry]] | OrderedDict[str, List[PluginEntry]]
+) = {}
 
 
 def check_type(image: ImageLike, reader_class: Reader) -> bool:
@@ -222,6 +226,15 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
             # Add plugin entry
             plugin_entry = PluginEntry(plugin, reader_meta, timestamp)
             for ext in plugin_entry.metadata.get_supported_extensions():
+                # Always remove leading "." if there is one
+                # Some plugin authors may add it, but it's not necessary
+                # And worse can cause this plugin list to have multiple entries
+                # for the same extension (e.g. ".tif" and "tif")
+                if ext.startswith("."):
+                    print(f"Plugin: {plugin_entry} includes leading '.' in ext: {ext}")
+                    ext = ext[1:]
+
+                # Start a new list of plugins for ext if it doesn't exist
                 if ext not in plugins_by_ext:
                     plugins_by_ext[ext] = [plugin_entry]
                     continue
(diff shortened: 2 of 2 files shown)


BrianWhitneyAI (approved):

bioio/plugins.py
```diff
@@ -233,6 +233,11 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
     # Save copy of plugins to cache then return
     plugins_by_ext_cache.clear()
     plugins_by_ext_cache.update(plugins_by_ext)
+
+    # Sort dict from longest to shortest key
+    plugins_by_ext = dict(
+        sorted(plugins_by_ext.items(), key=lambda x: len(x[0]), reverse=True)
+    )
```
toloudis, line 240: A few concerns about this.
1. you're changing the plugins_by_ext after its previous value was stored in the plugins_by_ext_cache.
2. this is completely undoing the "insert in sorted order by most recently installed" which we did above
3. maybe the least interesting of the three but it might have some implications -- this is a global sort over all file extensions right?  You're not just prioritizing ometiff over tiff, but really scrambling the order.

toloudis, line 240: A few concerns about this.
1. you're changing the plugins_by_ext after its previous value was stored in the plugins_by_ext_cache.
2. Is this completely undoing the "insert in sorted order by most recently installed" which we did above?
3. I'm not sure I understand how the sorting is happening. Is this a global sort over all extensions for all installed plugins? If so then you're not just prioritizing ometiff over tiff, but really scrambling the order.

evamaxfield: Completely agree with many points. This was my hacky fix before I needed to go before the end of the day and should have made this a draft PR.

Iirc our original goal was to sort the dict keys (file extensions) by longest first, and then within  each dict value (the list of plugin entries) sort the list of plugins by their install time.

The goal was basically both "always choose the most specific suffix and then within a suffix that has multiple supported readers, choose the most recently installed one"

I'll make changes tomorrow

evamaxfield: I can also do this processing at `determine_plugin` time if that's better?

evamaxfield:
Also resolves #56 by always _removing_ leading `.` from suffixes but happy to reverse that and always _add_ them instead.

evamaxfield:
Updates and I think ready for review!

Here is why the sorting matters, I am not changing any of the plugin lists in any way, only sorting the order in which each key-value pair is stored in the overall cache.

Prior to this change, here was the order of the plugin cache if you simply asked for the keys. This is the order in which readers would be tested in via [determine_plugin](https://github.com/bioio-devs/bioio/blob/main/bioio/bio_image.py#L164) (note that we are simply looping over the plugin cache items which defaults to _insert order_):

```
Before sorting:
['tif',
 'tiff',
 'ome.tiff',
 '264',
 '265',
 '3fr',
 '3g2',
 'A64',
 'IMT',
 'MCIDAS',
 'PCX',
 'SPIDER',
 'XVTHUMB',
 'a64',
 'adp',
 'amr',
 'amv',
 'apng',
 'arw',
 'asf',
 'avc',
 'avi',
 'avs',
 'avs2',
 'bay',
 'bif',
 'bmp',
 'cdg',
 'cgi',
 'cif',
 'ct',
 'dcr',
 'dib',
 'dip',
 'dng',
 'dnxhd',
 'dv',
 'dvd',
 'erf',
 'exr',
 'fff',
 'gif',
 'icb',
 'if',
 'iiq',
 'ism',
 'jif',
 'jfif',
 'jng',
 'jp2',
 'jpg',
 'mov',
 'mp4',
 'mpo',
 'msp',
 'pdf',
 'pngppm',
 'ps',
 'zif',
 'zarr']
```

After this change, here is the order of the plugin cache keys:

```
After sorting:
['ome.tiff',
 'XVTHUMB',
 'MCIDAS',
 'SPIDER',
 'pngppm',
 'dnxhd',
 'tiff',
 'apng',
 'avs2',
 'jfif',
 'zarr',
 'tif',
 '264',
 '265',
 '3fr',
 '3g2',
 'A64',
 'IMT',
 'PCX',
 'a64',
 'adp',
 'amr',
 'amv',
 'arw',
 'asf',
 'avc',
 'avi',
 'avs',
 'bay',
 'bif',
 'bmp',
 'cdg',
 'cgi',
 'cif',
 'dcr',
 'dib',
 'dip',
 'dng',
 'dvd',
 'erf',
 'exr',
 'fff',
 'gif',
 'icb',
 'iiq',
 'ism',
 'jif',
 'jng',
 'jp2',
 'jpg',
 'mov',
 'mp4',
 'mpo',
 'msp',
 'pdf',
 'zif',
 'ct',
 'dv',
 'if',
 'ps']
```

bioio/bio_image.py
```diff
@@ -164,9 +164,6 @@ def determine_plugin(
 
             # Check for extension in plugins_by_ext
             for format_ext, plugins in plugins_by_ext.items():
-                # Remove "." prefix if already there
-                if format_ext.startswith("."):
-                    format_ext = format_ext[1:]
```
evamaxfield, line 169: Moved to plugins.py as without it, duplicate suffixes are stored in the plugin cache

BrianWhitneyAI:
Is there an ome.tif plugin, dont see it on the list? I saw this error from a user `BioImage does not support the image: '/Users/r.dhamrongsirivadh/Desktop/test_images/LaminB1-Variance-multiChannels-2/AICS-13_2284.ome.tif'. You may need to install an extra format dependency. See bioio README for list of some known plugins.`

evamaxfield:
> Is there an ome.tif plugin, dont see it on the list? I saw this error from a user `BioImage does not support the image: '/Users/r.dhamrongsirivadh/Desktop/test_images/LaminB1-Variance-multiChannels-2/AICS-13_2284.ome.tif'. You may need to install an extra format dependency. See bioio README for list of some known plugins.`

There is an OME-TIFF plugin: https://github.com/bioio-devs/bioio-ome-tiff

That issue is what this PR resolves. It is also ready to review @bioio-devs/trusted-developers

BrianWhitneyAI:
> > Is there an ome.tif plugin, dont see it on the list? I saw this error from a user `BioImage does not support the image: '/Users/r.dhamrongsirivadh/Desktop/test_images/LaminB1-Variance-multiChannels-2/AICS-13_2284.ome.tif'. You may need to install an extra format dependency. See bioio README for list of some known plugins.`
> 
> There is an OME-TIFF plugin: https://github.com/bioio-devs/bioio-ome-tiff
> 
> That issue (if they have the ome-tiff plugin installed) is what this PR resolves. It is also ready to review @bioio-devs/trusted-developers

This is with bioio-ome-tiff installed I believe

toloudis:
@SeanLeRoy I kinda want to find some of our discussions about why we chose the simpler install-order over this, to make sure we are not missing some good reason for something.  I almost think it would more conservative and a less dramatic change if you did this length-sorting only for when there's a "ext2.ext" situation, like keep the order the same except for making sure ome.tiff comes before tiff and ome.tif comes before tif.  Are there any other extensions we know of, that this applies to?

evamaxfield:
> @SeanLeRoy I kinda want to find some of our discussions about why we chose the simpler install-order over this, to make sure we are not missing some good reason for something. I almost think it would more conservative and a less dramatic change if you did this length-sorting only for when there's a "ext2.ext" situation, like keep the order the same except for making sure ome.tiff comes before tiff and ome.tif comes before tif. Are there any other extensions we know of, that this applies to?

Again, to be clear. This isn't _directly_ changing the install order. Each plugin list is still in install order, this just determines which plugin list to try first.

bioio/plugins.py
```diff
@@ -37,7 +38,9 @@ class PluginEntry(NamedTuple):
 
 
 # global cache of plugins
-plugins_by_ext_cache: Dict[str, List[PluginEntry]] = {}
+plugins_by_ext_cache: Union[
+    Dict[str, List[PluginEntry]], OrderedDict[str, List[PluginEntry]]
```
toloudis, line 42: should only be OrderedDict, right?

bioio/plugins.py
```diff
@@ -222,6 +225,14 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
             # Add plugin entry
             plugin_entry = PluginEntry(plugin, reader_meta, timestamp)
             for ext in plugin_entry.metadata.get_supported_extensions():
+                # Always remove leading "." if there is one
+                # Some plugin authors may add it, but it's not necessary
+                # And worse can cause this plugin list to have multiple entries
+                # for the same extension (e.g. ".tif" and "tif")
+                if ext.startswith("."):
```
toloudis, line 232: I can't remember if our base Reader docs suggest that extension strings start with dot or not, but this change seems good.

bioio/bio_image.py
toloudis: this one-by-one linear search kinda sucks but I don't have any better ideas for how to do it (the following line after this change).  I guess generally it's better to just specify the Reader up front :)

toloudis (approved):
one comment on the Dict/OrderedDict typing that should be simplified but overall LGTM!

bioio/bio_image.py
evamaxfield: As a nitpick I would also like to push for having it always _add_ the `.` at the front rather than remove it. As we are already doing that a line below with:

`if path.lower().endswith(f".{format_ext}"):`

This line is also making me realize that we need to lowercase all of the ext in the plugin cache as well (there are already plugins (bioio-imageio) which are reporting capitalized suffixes).

I can do those two things in another PR. Just would love your quick thoughts. It just seems like we should be doing as much of the extension prep and manipulation in the plugin cache creation rather than at `determine_plugin` time.

bioio/plugins.py
evamaxfield: I believe you are correct but might need to make the function have a plugins_by_ext_unordered variable in addition to appease mypy. Will give it a whirl.

bioio/plugins.py
evamaxfield: I would like to make even more changes to this but happy to do so in another PR: https://github.com/bioio-devs/bioio/pull/55/files#r1688696186

evamaxfield (commented):
@BrianWhitneyAI here were the things I wanted to do before this gets merged if you want to pick this up for me. Sorry got busy with a paper deadline, friends staying with us, and packing for a move all at the same time 😢.

1. In `bioio/bio_image.py` there is a `if path.lower().endswith()...`. I agree that that `lower()` is important and good but we should lowercase all of the plugin suffixes during plugin cache creation. So simply add a `lower()` to the suffix before it goes into the cache.

2. In `bioio/plugins.py` I updated the code to always check for leading `"."` and remove it if found. I think this should be swapped to checking if the first character _isn't_ a `"."` and adding one if it isn't found. My reasoning for this is that it's entirely possible that a filepath ends in "ome.tiff" without it being an actual ".ome.tiff" for example the "4d-nucleome" project (if that is still going on) may have a "4d-nucleome.tiff" that would be detected as an "ome.tiff" under the current "ends with "ome.tiff" and not "ends with .ome.tiff".

3. @toloudis asked me to check if the typing of the plugin cache can be fully swapped to only `OrderedDict`. Currently the typing is either `dict[str, list[PluginEntry]]` and it may be possible to only have it as a single type.

Hope that all makes sense!!

BrianWhitneyAI:
@evamaxfield  hopefully resolved all of your points in [d470d7f](https://github.com/bioio-devs/bioio/pull/55/commits/d470d7f17437f8785b5d4aa4edacb3ad81d982c2)!

evamaxfield (commented):
Thanks @BrianWhitneyAI all looks good to me!! Thanks for finishing this work ❤️

PR #67, opened 2024-08-15
Support public s3 files automatically

### LINKED ISSUE #61: Read from public `s3://` paths without authentication, without requiring any code change from users
### Feature Description

If a file is hosted publicly on S3 and a user without AWS credentials set up must use `fs_kwargs`: `BioImage("s3://bucketname/path/to/file", fs_kwargs=dict(anon=True))`. (I'm thinking specifically about OME ZARRs, but this is likely relevant to all readers.)

Instead, bioio should be able to handle this internally and let the user write `BioImage("s3://bucketname/path/to/file")`.

### Solution

As far as I can tell, the proper way to check if a user is authenticated to read a file is to attempt to read it and see if there's an error, so the solution I think is to try to read files twice with logic similar to the following.
```python
try:
   # __init__ with user's fs_kwargs
except SomethingSpecific as e:
   if protocol == "s3://":
       # __init__ with user's fs_kwargs plus {anon: True}
   else:
       raise e
``` 

### Alternatives

Looks like they [tried it in s3fs](https://github.com/fsspec/s3fs/pull/823) but was reverted “unfortunately, it led to far more problems than it solved. I’d be happy to see a more solid implementation, if some wants to try.”

SeanLeRoy:
### Link to Relevant Issue

This pull request resolves #61 

### Description of Changes

When reading from an S3 protocol our current file system handlers require either explicit credentials passed in or an explicit declaration of an "anonymous" reader. This change makes it so we try to read a file given the current `fs_kwargs` (used by the readers for S3) and if that fails we try to read again as an explicitly "anonymous" reader.

**Additionally** I made it so we only create a text file once and only install the dummy-plugin once because I felt running the tests was rather slow and this sped it up IMO.

### Testing
~Added a unit test, unable to get this to work the S3 file I was provided `s3://allencell/aics/emt_timelapse_dataset/data/3500005548_25_all_cells_mask.ome.zarr`~

bioio-ome-zarr has unreleased changes, when using the newest version of bioio-ome-zarr that is unreleased it works, so I am working on releasing that now but should be GTG after that

### DIFF
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -255,6 +255,35 @@ class BioImage(biob.image_container.ImageContainer):
             ),
         )
 
+    @staticmethod
+    def _get_reader(
+        image: biob.types.ImageLike,
+        reader: biob.reader.Reader,
+        use_plugin_cache: bool,
+        fs_kwargs: Dict[str, Any],
+        **kwargs: Any,
+    ) -> Tuple[biob.reader.Reader, Optional[PluginEntry]]:
+        """
+        Initializes and returns the reader (and plugin if relevant) for the provided
+        image based on provided args and/or the available bioio supported plugins
+        """
+        if reader is not None:
+            # Check specific reader image types in a situation where a specified reader
+            # only supports some of the ImageLike types.
+            if not check_type(image, reader):
+                raise biob.exceptions.UnsupportedFileFormatError(
+                    reader.__name__, str(type(image))
+                )
+
+            return reader(image, fs_kwargs=fs_kwargs, **kwargs), None
+
+        # Determine reader class based on available plugins
+        plugin = BioImage.determine_plugin(
+            image, fs_kwargs=fs_kwargs, use_plugin_cache=use_plugin_cache, **kwargs
+        )
+        ReaderClass = plugin.metadata.get_reader()
+        return ReaderClass(image, fs_kwargs=fs_kwargs, **kwargs), plugin
+
     def __init__(
         self,
         image: biob.types.ImageLike,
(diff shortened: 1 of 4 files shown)


bioio/tests/conftest.py
```diff
@@ -36,10 +36,12 @@
...
 
 @pytest.fixture
-def sample_text_file(tmp_path: pathlib.Path) -> pathlib.Path:
+def sample_text_file(
+    tmp_path: pathlib.Path,
+) -> typing.Generator[pathlib.Path, None, None]:
     example_file = tmp_path / "temp-example.txt"
     example_file.write_text("just some example text here")
-    return example_file
+    yield example_file
```
SeanLeRoy, line 44: Unrelated to main changeset, this is just to make the tests faster by only creating the file once since they seemed slow

bioio/tests/conftest.py
```diff
@@ -90,3 +92,9 @@ def __exit__(
         subprocess.check_call(
             [sys.executable, "-m", "pip", "uninstall", "-y", self.package_name]
         )
+
+
+@pytest.fixture
+def dummy_plugin() -> typing.Generator[str, None, None]:
+    with InstallPackage(package_path=DUMMY_PLUGIN_PATH, package_name=DUMMY_PLUGIN_NAME):
+        yield DUMMY_PLUGIN_NAME
```
SeanLeRoy, line 100: Unrelated to main changeset, this is just to make the tests faster by only installing the plugin once since they seemed slow

bioio/bio_image.py
```diff
@@ -264,25 +293,28 @@ def __init__(
...
+            # try reading as an anonymous user otherwise re-raise
+            # the error
+            if not str(image).startswith("s3://"):
+                raise
+
+            self._reader, self._plugin = self._get_reader(
+                image,
+                reader,
+                use_plugin_cache,
+                fs_kwargs | {"anon": True},
```
toloudis, line 315: the implication here, then, is that if you passed fs_kwargs={anon:True} in the initial bioImage constructor then s3 reads do work?

SeanLeRoy: Yeah 👍 (and I believe were we to always passed `True` here then users that need to auth like for private buckets would fail, haven't tested that explicitly yet)

toloudis (approved):
Looks good!  I do wonder if we can handle the exception at an even lower level to avoid needing to run determine_plugin twice, but it's probably not a big deal for performance.

SeanLeRoy:
`bioio-ome-zarr` at version `1.1.0` has been released and works with this :) it seems the publish workflow for `bioio-ome-zarr` was broken from this change https://github.com/bioio-devs/bioio-ome-zarr/pull/26 so I fixed it with https://github.com/bioio-devs/bioio-ome-zarr/commit/0f3976d17585eab3b7a723e0cb090faa53f4d9e9 and now we are all good

SeanLeRoy:
> Looks good! I do wonder if we can handle the exception at an even lower level to avoid needing to run determine_plugin twice, but it's probably not a big deal for performance.

Yeah I couldn't find a better place other than inside each individual plug-in and I thought it would be nicer to take the hit here for S3 files rather than be forced to do in in each plugin. Totally down to try something else out though if someone has an idea of where this could go it does seem like a performance bummer to have to straight up retry the read.

toloudis:
> Yeah I couldn't find a better place other than inside each individual plug-in and I thought it would be nicer to take the hit here for S3 files rather than be forced to do in in each plugin. Totally down to try something else out though if someone has an idea of where this could go it does seem like a performance bummer to have to straight up retry the read.

one thought: Can we tell if we need the anon key by checking early for "s3://" in the path and looking at what fs_kwargs were passed in?  If anon is omitted, I assume the s3fs just looks for environment variables with credentials or aws credential files. So maybe it's hard for us to make a good guess.

BrianWhitneyAI (approved):
Looks good to me! Thanks for fixing up the dummy reader stuff :)

SeanLeRoy:
> I assume the s3fs just looks for environment variables with credentials or aws credential files. So maybe it's hard for us to make a good guess.

I believe this is true, I think it looks at the same cred file `s3` CLI and `boto3` look at under `/.aws/credentials`

pgarrison:
Thanks @SeanLeRoy! So glad to see this taken care of

Issue #53, opened 2024-07-17
remove zarr<2.18.0 constraint


toloudis:
### Feature Description

Tests were failing on #48 because of more recent zarr versions so I had to pin to pre-2.18.  Ideally this should not be a constraint.

### Use Case

We want to be able to take advantage of all the fixes and improvements that come with new zarr releases!

### Solution

Unpin to zarr<3 and eventually allow v3 support too

SeanLeRoy:
Resolved as of now

PR #72, opened 2024-09-03
unpin zarr max version

joshua-gould:


evamaxfield (approved):
Fine with me!

For reference for others, I also had to make this change locally when working on the previous PRs with bioio-ome-zarr and bioio packages to get them to work together. `bioio-ome-zarr` has a pin above this upper bound: https://github.com/bioio-devs/bioio-ome-zarr/blob/main/pyproject.toml#L32 so they would conflict on install of both.

evamaxfield (changes requested):
Prior review was just to say that I am fine with us trying to unpin the upper bound but the tests need to be fixed... :(

toloudis:
I've seen this error before but can't remember how it was fixed.  Somehow the tests were passing when it was merged.

joshua-gould:
Is there a reason, we can't use https://ome-zarr.readthedocs.io/en/stable/api/writer.html#ome_zarr.writer.write_multiscale to write? Thanks.

toloudis:
Sorry, here is the weird issue I was having before and the reason why it was pinned at all:  https://github.com/bioio-devs/bioio/issues/53  
this describes the exact unit test failure we see here.  We will have to dig to figure out why this is a problem. I'd love to get it fixed.

toloudis:
You absolutely could use  https://ome-zarr.readthedocs.io/en/stable/api/writer.html#ome_zarr.writer.write_multiscale .  We had trouble with performance and scalability on some of our larger data set conversions.  The writer included here is set up to convert long time series images at AICS. It is going to evolve as we generalize to more and more varied data sets. We would love to arrive at a robust and scalable general zarr writer that has a convenient and useful API.  Anyone is free to use the tools in ome-zarr-py without using the Writers in here of course.

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -452,7 +452,7 @@ def _downsample_and_write_batch_t(
         data_tczyx = data_tczyx.persist()
         # data_tczyx.compute()
         for k in range(start_t, end_t):
-            self.levels[0].zarray[k] = data_tczyx[k - start_t]
+            self.levels[0].zarray[k] = data_tczyx[k - start_t].compute()
```
toloudis, line 455: I was wondering if the compute call here is redundant with the persist call above, but it seemed to fix the unit test issue?

BrianWhitneyAI (approved):

bioio/writers/ome_zarr_writer_2.py
```diff
@@ -449,32 +449,21 @@ def _downsample_and_write_batch_t(
...
-            # lvls[j].set_basic_selection(slice(start_t,end_t), ti[:])
             for k in range(start_t, end_t):
-                self.levels[j].zarray[k] = data_tczyx[k - start_t]
-            # for some reason this is not working: not allowed to write in
-            # this way to a non-memory store
-            # dask.array.to_zarr(ti, lvls[j], component=None,
-            # storage_options=None, overwrite=False, region=(slice(start_t,
-            # end_t)))
+                subset = data_tczyx[[k - start_t]]
+                da.to_zarr(subset, self.levels[j].zarray, region=(slice(k, k + 1),))
```
toloudis, line 466: This is looking awesome. My only concern with these changes is that they are doing the compute step once for every T time rather than in a batch of times.  For example, I think it would be more equivalent if you did da.to_zarr with a region that had all the t values from start_t to end_t in it.  I am being a little cautious about this because this is a core piece of code we use to do our conversions and it will be a while before I could really test it.  Honestly at this point it does look good to me.  I'm happy that it fixes the problem with zarr version.  
 @vianamp Do you have an opinion on this code change?

joshua-gould: Another option is to pass compute=False to da.to_zarr and then call compute once on all the delayed objects

vianamp: This looks good to me.

vianamp (approved):

toloudis (approved):

toloudis:
@joshua-gould thank you so much for this contribution.
@evamaxfield ok to merge this??

evamaxfield (approved):
This all looks good to me! Thank you @joshua-gould ❤️ 

@toloudis I second your concern / I am also unsure how many times compute is going to be called on individual T-slices. That said, I think this is a ship it and test it out situation rather than hold this back. `bioio` and `bioio-ome-zarr` conflicting is potentially a larger issue.

@vianamp if you have any datasets that you are actively processing, one way to test this would be to run the processing script with current `bioio` version and track peak memory usage + CPU time and then upgrade and track memory and CPU again to compare??

Regardless, merging!

PR #74, opened 2024-09-05
Update ome_tiff_writer.py to remove reference to `image_names`

TimMonko:
### Link to Relevant Issue

This pull request resolves #

### Description of Changes

Documentation suggests that argument is `image_names` instead of `image_name`. Using `OmeTiffWriter.save(image_names=[str])` silently saves the data without saving the argument to the ome metadata. This fixes updates the documentation so users instead use `OmeTiffWriter.save(image_name=[str])` to avoid this issue.


evamaxfield (approved):
Thank you!! These always annoy me too.

BrianWhitneyAI (approved):

Issue #78, opened 2024-12-06
`ngff-zarr==0.10.0` breaks bioio


TimMonko:
### Describe the Bug
`ngff-zarr` released an update to [0.10.0 two days ago](https://github.com/thewtex/ngff-zarr/releases/tag/v0.10.0). 

In [PR#109](https://github.com/thewtex/ngff-zarr/pull/109) the `ngff-zarr.zarr_metadata` module is renamed to `ngff-zarr.v04.zarr_metadata`. Not sure the motivation to rename the module (besides some support for v05 that I'm not reading in to). But, I assume your CI is about to break (as did my own tests).

Pinning `ngff-zarr<0.10.0` solves the issue. 

e.g.:
```
  .tox/py310-linux/lib/python3.10/site-packages/bioio/writers/ome_zarr_writer_2.py:11: in <module>
      from ngff_zarr.zarr_metadata import Axis, Dataset, Metadata, Scale, Translation
  E   ModuleNotFoundError: No module named 'ngff_zarr.zarr_metadata'
```

kmitcham:
This issue seems like it was resolved in 
https://github.com/thewtex/ngff-zarr/releases/tag/v0.10.1
The only version that would be a problem is EXACTLY v0.10.0, which is neither the min nor the latest ([ngff-zarr 0.12.1](https://github.com/thewtex/ngff-zarr/releases/tag/v0.12.1) )

PR #80, opened 2025-01-08
Switch from ngff-zarr to ome-zarr-models

imagejan:
### Link to Relevant Issue

See https://github.com/bioio-devs/bioio-base/issues/25#issuecomment-2575338781, as well as https://github.com/bioio-devs/bioio-ome-zarr/issues/29.

This PR also makes #78 obsolete.

### Description of Changes

The `ngff-zarr` dependency is replaced by `ome-zarr-models`, mapping the same metadata classes of the OME-NGFF spec.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.9, "3.10", 3.11, 3.12]
+        python-version: [3.11, 3.12, 3.13]
         os: [ubuntu-latest, macos-latest, macos-13, windows-latest]
 
     steps:
--- a/bioio/writers/ome_zarr_writer_2.py
+++ b/bioio/writers/ome_zarr_writer_2.py
@@ -1,5 +1,5 @@
 import logging
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass
 from math import prod
 from typing import Any, List, Tuple, Union
 
@@ -8,7 +8,12 @@ import numcodecs
 import numpy as np
 import skimage.transform
 import zarr
-from ngff_zarr.zarr_metadata import Axis, Dataset, Metadata, Scale, Translation
+from ome_zarr_models.v04.axes import Axis
+from ome_zarr_models.v04.coordinate_transformations import (
+    VectorScale,
+    VectorTranslation,
+)
+from ome_zarr_models.v04.multiscales import Dataset, Multiscale
 from zarr.storage import DirectoryStore, FSStore, default_compressor
 
 from bioio import BioImage
@@ -646,19 +651,23 @@ class OmeZarrWriter:
                 # TODO handle optional translations e.g. xy stage position,
                 # start time etc
                 translation.append(0.0)
-            coordinateTransformations = [Scale(scale), Translation(translation)]
+            coordinateTransformations = (
+                VectorScale(type="scale", scale=scale),
+                VectorTranslation(type="translation", translation=translation),
+            )
             dataset = Dataset(
                 path=path, coordinateTransformations=coordinateTransformations
             )
             datasets.append(dataset)
-        metadata = Metadata(
+        metadata = Multiscale(
+            version=OME_NGFF_VERSION,
             axes=axes,
             datasets=datasets,
             name="/",
             coordinateTransformations=None,
         )
 
-        metadata_dict = asdict(metadata)
+        metadata_dict = metadata.model_dump()
         metadata_dict = _pop_metadata_optionals(metadata_dict)
 
         # get the total shape as dict:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,9 +22,9 @@ authors = [
 classifiers = [
   "Development Status :: 4 - Beta",
   "Natural Language :: English",
-  "Programming Language :: Python :: 3.9",
-  "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
 ]
 dynamic = ["version"]
 dependencies = [
@@ -32,10 +32,10 @@ dependencies = [
   "dask[array]>=2021.4.1",
   "fsspec>=2022.8.0",
   "imageio[ffmpeg]>=2.11.0,<2.28.0",
-  "ngff-zarr>=0.8.2",
   "numpy>=1.21.0,<2.0.0",
   "ome-types[lxml]>=0.4.0",
   "ome-zarr>=0.6.1",
+  "ome-zarr-models>=0.1.1",
   # This version was revoked from PyPi, but in case of using a stale version
   # or a private package index, this will prevent usage
   "scikit-image!=0.23.0",


toloudis:
Code changes look good, nice to see that it was a pretty simple change.

I am not super familiar with the ome-zarr-models repo yet.  
One of the considerations here needs to be future versions of ome-zarr 0.5 and above.  Any idea how soon this library or the Janelia pydantic one will have full implementation of 0.5?  Do the maintainers of ome-zarr-models want to be aggressive about supporting published ome-ngff specs? 

Since this code is used by our ome-zarr writer, we want, at least internally, to be ready to deal with sharded zarrv3 some time soon (vague, I know).

toloudis:
I just found context around the choice of package over here: https://github.com/bioio-devs/bioio-base/issues/25#issuecomment-2577117239

imagejan:
> Any idea how soon this library or the Janelia pydantic one will have full implementation of 0.5? Do the maintainers of ome-zarr-models want to be aggressive about supporting published ome-ngff specs?

`ome-zarr-models-py` aims to be a reference implementation for the NGFF spec.
See https://ome-zarr-models-py.readthedocs.io/en/latest/#design

Also, RFC authors are encouraged to implement proposed changes directly in `ome-zarr-models-py` to test them in practice, so I propose this would be the package of choice to support any new version of the spec.
See https://ome-zarr-models-py.readthedocs.io/en/latest/contributing/#implementing-ome-zarr-rfcs

> Since this code is used by our ome-zarr writer, we want, at least internally, to be ready to deal with sharded zarrv3 some time soon (vague, I know).

Fully agreed with this goal 🙂 

How strong is the need to support Python 3.9 and 3.10 in `bioio`?

SeanLeRoy (commented):
Code changes seem good to me too! Does this drop support for 3.10/3.11 or was that just a side question @imagejan?

imagejan:
> Does this drop support for 3.10/3.11 or was that just a side question

Yes, unfortunately, `ome-zarr-models` only support Python 3.11+, so I assume this might be a blocker?!

(My motivation for removing ngff-zarr is to simplify deployment of `bioio` to conda-forge.)

toloudis:
As much as I support this PR, I think supporting 3.9 and 3.10 is a requirement to continue to use this within AICS for now.  We have to update other computational python code before we can restrict bioio in that way.

If this code change removes support for 3.9 and 3.10 then we have to modify pyproject.toml and the unit tests and possibly consider it a major version change, or say it very loudly in release notes..

We will have to decide how best to move forward!

evamaxfield:
> As much as I support this PR, I think supporting 3.9 and 3.10 is a requirement to continue to use this within AICS for now. We have to update other computational python code before we can restrict bioio in that way.
> 
> If this code change removes support for 3.9 and 3.10 then we have to modify pyproject.toml and the unit tests and possibly consider it a major version change, or say it very loudly in release notes..
> 
> We will have to decide how best to move forward!

Is there anything coming down the pipeline for this repo / plugins that AICS internal is looking to use? If there isn't anything that needs upgrading in BioIO for AICS internal usage, I don't necessarily see the problem in allowing this PR to go through (and bumping the minimum supported versions) if there aren't features or bugfixes needed from AICS internally because everyone should be able to continue using the various BioIO versions already in use.

That said, if it is a case of waiting for current features and bugfixes to go through before updating min supported versions, here is the list of PRs open in the overarching project:

## Maybe AICS Internal?
- https://github.com/bioio-devs/bioio-base/pull/24 -- should be able to be merged, just needs a branch update to get up-to-date with main and CI checkoff
- https://github.com/bioio-devs/bioio-base/pull/20 -- I think good to go?

## Maybe AICS External?
- https://github.com/bioio-devs/bioio-czi/pull/18 -- issue from pyarrow conflict with bioio (I still believe it to be related to an upstream but unclear)
- https://github.com/bioio-devs/bioio-imageio/pull/11/files -- placing this in external as I don't think bioio is used for PNG reading internally but it is a bug that is longstanding and I am likely the holdup on this
- https://github.com/bioio-devs/bioio-bioformats/pull/17 and https://github.com/bioio-devs/bioio-bioformats/pull/18 -- seem related to each other, and larger issue with underlying bioformats changes

toloudis:
We have one or two major internal computational packages that need updating to 3.11+. 

Future key work for bioio:

* Internal: Hard to predict right now but there are lots of anticipated issues IF we ramp up our usage of non-zeiss imaging. (e.g. nd2, sldy, lif)

* zarr v3 adoption is internal+external impact. Unclear when we will start on it but the sooner we start experimenting the better, because it affects storage efficiency.

* Also we have a long standing issue with the way we read chunks only at whole-slice granularity, if I am remembering right.  Which affects dask efficiency and the ability to read the same chunk sizes that we will write back to zarr.  I think most of our zarr conversions could incur a rechunking on dask array.

imagejan:
FWIW, I changed the python support in this PR to `>=3.11, <=3.13`. Feel free to close (or leave around unmerged) if this doesn't fit the current plans for `bioio`.

I'd be happy with any route towards making `bioio` and all its dependencies available on conda-forge.

toloudis:
I guess I could be on board with:
* basically version every package in bioio-devs at something like 1.99 (silly example number)
* then cut a 2.0 that kills the older python support.  Announce what we did.
* move on with our lives
BUT: I'm fairly sure that if AICS internal is stuck on old python, we will still need improvements to some of the bioio readers and make a zarrv3 writer AT LEAST.
@pgarrison is there some kind of path to drag AICS internal python code up to >=3.11?

pgarrison:
> @pgarrison is there some kind of path to drag AICS internal python code up to >=3.11?

With the exception of [cyto-dl](https://github.com/AllenCellModeling/cyto-dl/) and current projects that use `cyto-dl`, AFAIK most of our code that is stuck on Python < 3.11 is legacy code that probably doesn't need to be upgraded to the latest `bioio` version. 

Even if bioio 2 drops support for Python < 3.11, we could consider backporting some bugfixes to bioio 1.x as necessary, though I know we'd much rather have a single version to maintain.

toloudis:
After one more round of internal discussions, personally I feel totally comfortable with moving ahead with this and basically following (or keeping up with) the scientific-python roadmap.  I'm all in favor of it.  I would just like us to be careful about communicating what is the last bioio version before changing the supported python version.

toloudis (approved):
LGTM, maybe a followup pr for documentation about python version etc before we do a release

kmitcham (approved):

toloudis:
can merge?

SeanLeRoy:
> can merge?

Don't we need to do a release for 3.9/3.10 before dropping them?

kmitcham:
> > can merge?
> 
> Don't we need to do a release for 3.9/3.10 before dropping them?

https://github.com/bioio-devs/bioio/releases/tag/v1.3.0 drops support for 3.9 officially.

BrianWhitneyAI:
@imagejan This is good to go after the 3.9 updates in https://github.com/bioio-devs/bioio/pull/105 correct? Will also need to resolve some conflicts. Thanks again!

BrianWhitneyAI:
added in [601132f](https://github.com/bioio-devs/bioio/commit/601132f8ebaa9ee633e87db74838652463ae8599)

PR #88, opened 2025-01-30
feature/parameterized-ome-tiff-writer-compression

### LINKED ISSUE #36: writer compression parameter
### Feature Description

It may benefit many users to be able to set compression parameters for both zarr and tiff writing.

### Use Case

See https://github.com/AllenCellModeling/aicsimageio/issues/553

### Solution

Add compression args to writer classes.  I guess these can be specific to the supported compression types of the underlying libraries.  Consider just forwarding kwargs through.

### LINKED ISSUE #87: [Feature Request] Image compression for ome.tiff files
### Feature Description

Include the ability to use image compression when saving `ome.tiff` files by providing additional input arguments in `OmeTiffWriter().save()` that are passed on to `tifffile.imsave()` or `tifffile.imwrite()`. This feature could also be implemented for other file formats if possible.

### Use Case

Using image compression to save image segmentation or intermediate images used for analysis could help to reduce storage space on hard drives and network storage. Segmentation images are often sparse and comprised of large regions of identical pixel values, making them easy to compress to a substantial degree with minimal loss. Likewise raw images can also be compressed in scenarios where loss is a negligible concern. 

### Solution

Image compression parameters [already exist](https://github.com/cgohlke/tifffile/blob/016c474c3a28ae08968ffa9e24efcbf0dbdda8b1/tifffile/tifffile.py#L1316) in `tifffile`, which `OmeTiffWriter()` wraps. We need only provide these same arguments as optional parameters to `OmeTiffWriter().save()` that are passed down to `tifffile.imwrite()` when writing the image to disk, and include relevant documentation explaining options to users. Additional messaging to users about this feature to help in adopting it as standard practice can be discussed.

evamaxfield:
…ME TIFF Writer

### Link to Relevant Issue

This pull request:
- Resolves #87
- Resolves #36 

### Description of Changes

This moves the hard coded compression setting in OmeTiffWriter to a parameter.

My only general discussion point before this gets merged is perhaps maybe we should just open up all tifffile kwargs like how we do with `fs_kwargs`?


toloudis:
It says it resolves #36 but I think that one mentions both tiff and zarr.  We need to at least have another open issue for zarr.

bioio/writers/ome_tiff_writer.py
```diff
@@ -46,6 +47,7 @@ def save(
             Union[List[List[int]], List[Optional[List[List[int]]]]]
         ] = None,
         fs_kwargs: Dict[str, Any] = {},
+        compression: enum.IntEnum = TIFF.COMPRESSION.ADOBE_DEFLATE,
```
toloudis, line 50: compression options are missing

toloudis: maybe enough to just say something in the docstring about them and rely on kwargs?

evamaxfield (commented):
@toloudis this PR has been updated to allow for all tifffile write kwargs to be passed. This should resolve the lack of compression args and generally just future proofing us a bit in the case people want to use other parameters.

Docstring includes custom compression example

evamaxfield:
Quick note: seems like `TIFF.COMPRESSION` and the like have been deprecated since 2022 with preference for using `tifffile.COMPRESSION` etc. Do we want to make that change? I guess people could theoretically still be using tifffile 2021 versions but 🤷‍♀️

toloudis:
I'm generally in favor of using current things but we might have to put a min value on our tifffile version?

evamaxfield:
Ping on this. I don't think we need to change our minimum tifffile version for a bit.

Good to merge though? @toloudis @SeanLeRoy @BrianWhitneyAI

toloudis:
> Ping on this. I don't think we need to change our minimum tifffile version for a bit.
> 

I thought you were going to do something to use the more current (post-2022) enum

toloudis (approved):
LGTM, thanks!

SeanLeRoy (approved):

PR #89, opened 2025-02-04
restict ngff-zarr to prevent errors in Ubuntu 18

kmitcham:
### (resolved before issue created)
Chantelle was encountering issues with bioio writer imports when using latest versions of bio.
We eventually traced the issue to Ubuntu 18 having issues with ngff-zarr>0.9.0, specifically its dependencies in 
itkwasm and wasmtime.  I don't think I need to pin the wasm dependencies, but if we do, the versions are 
itkwasm==1.0b171 wasmtime==22 
pip install ngff-zarr==0.9.0 itkwasm==1.0b171 wasmtime==22 

### Description of Changes
Put a maximum version number on ngff-zarr to avoid low level lib errors.


evamaxfield (commented):
I am fine with this but I believe #80 will also fix this problem

toloudis (commented):
LGTM?  I also agree about #80 but this is a good enough interim thing

toloudis (approved):

BrianWhitneyAI (approved):

Issue #96, opened 2025-03-06
Build failing for python 3.10 and 3.11


kmitcham:
### Describe the Bug

CI builds are failing.   This appears to be due to a new version of tiffile changing the names of some constants.
2024.12.12 carried the warning 
DeprecationWarning: <tifffile.TIFF.COMPRESSION> is deprecated since 2022.7.28. Use tifffile.COMPRESSION
and 2025.2.18 has the error:
    DEFAULT_TIFF_WRITE_KWARGS = {"compression": TIFF.COMPRESSION.ADOBE_DEFLATE}
E   AttributeError: '_TIFF' object has no attribute 'COMPRESSION'. Did you mean: 'COMPRESSORS'?

### Expected Behavior

bioio should pass all tests in python 3.10 and 3.11

### Reproduction

run 'just test' on main.

I think is is another issue where dropping 3.9 makes the fix much simpler, and hence is gated on that decision being formalized.
A hacky work-around would be to pin the tifffile version to < 2025.2.18

karthi03857:
The issue is due to TIFF.COMPRESSION being removed in tifffile 2025.2.18.
To fix it

import tifffile
DEFAULT_TIFF_WRITE_KWARGS = {"compression": tifffile.COMPRESSION.ADOBE_DEFLATE}

If an immediate fix is not possible, pin tifffile to an older version in requirements.txt:

tifffile<2025.2.18

Then reinstall dependencies:

pip install -r requirements.txt

kmitcham:
Spoke with Sean, and we will be dropping 3.9, adding 3.12, so a simple code update as per Karthi is what I will be doing.  Then I get to update the CI process to match that version change.

PR #97, opened 2025-03-06
update to use new tifffile enumerations

kmitcham:
This fix drops support for python 3.9, but allows use of the latest tifffile library.

### Link to Relevant Issue
https://github.com/bioio-devs/bioio/issues/96

### Description of Changes

Update reference location to some enumerations


kmitcham:
bad branch name.  Will re-open with properly formatted branch name.
Automatic branch creation does not work with our desired format.

PR #98, opened 2025-03-06
Feature/96 update constant name

kmitcham:
This fix drops support for python 3.9, but allows use of the latest tifffile library.

### Link to Relevant Issue
https://github.com/bioio-devs/bioio/issues/96

### Description of Changes

Update reference location to some enumerations.  Update CI to not run 3.9

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.9, "3.10", 3.11, 3.12]
+        python-version: ["3.10", 3.11, 3.12]
         os: [ubuntu-latest, macos-latest, macos-13, windows-latest]
 
     steps:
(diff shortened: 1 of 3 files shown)


toloudis (approved):
I'm good... does this mean we need to issue a versioned release first (that would be our last python 3.9 release)? 
Also does this require a new min version pin for tifffile?

Lastly, can we add 3.13 safely yet? (maybe for a different PR)

kmitcham:
> I'm good with this... Does this mean we need to issue a versioned release first (that would be our last python 3.9 release)? Also does this require a new min version pin for tifffile?
> 
I'm not sure what the release process is in terms of deprecating stuff, or more in general, actually.  

> Lastly, can we add 3.13 safely yet? (maybe for a different PR)
I think that might be broad enough for another PR; CI is currently only going up to 3.12

SeanLeRoy:
>Does this mean we need to issue a versioned release first (that would be our last python 3.9 release)?

Yes for sure! Lets do that then merge this then add Python 3.13 if possible in another PR.

SeanLeRoy (approved):

BrianWhitneyAI (approved):

kmitcham:
Did a release for the last 3.9 version

PR #105, opened 2025-03-26
Clarify supported python versions

imagejan:
This is a follow-up on https://github.com/bioio-devs/bioio/commit/c5bdbff91124f21775eed01b0200bbb4619c217c, removing Python 3.9 also from the metadata, and adding Python 3.13 to the CI build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", 3.11, 3.12]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
         os: [ubuntu-latest, macos-latest, macos-13, windows-latest]
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", 3.11, 3.12]
+        python-version: [3.10, 3.11, 3.12, 3.13]
```
pgarrison, line 40: The quotes around 3.10 are required: without them it is interpreted as 3.1

pgarrison (changes requested):
Thanks for submitting this PR!

BrianWhitneyAI (changes requested):
Thanks for submitting this! also could go for a readme update!

.github/workflows/ci.yml
BrianWhitneyAI: yeah trailing zero breaks unfortunately

BrianWhitneyAI:
resolved in [601132f](https://github.com/bioio-devs/bioio/commit/601132f8ebaa9ee633e87db74838652463ae8599)

PR #110, opened 2025-04-02
feature/standard-metadata

BrianWhitneyAI:
### Description 

This pr is a companion PR to https://github.com/bioio-devs/bioio-base/pull/29 adding the standard_metadata property to BioImage so that it can be accessed by top level users.

(tests will fail until base is released adding ImageContainer.standard_metadata)

### DIFF
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -1003,6 +1003,14 @@ class BioImage(biob.image_container.ImageContainer):
         """
         return self.reader.time_interval
 
+    @property
+    def standard_metadata(self) -> biob.standard_metadata.StandardMetadata:
+        """
+        Return the manifest of standard metadata as a dictionary. The possible
+        fields are predefined by the StandardMetadata dataclass.
+        """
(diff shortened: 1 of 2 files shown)


bioio/bio_image.py
```diff
@@ -1003,6 +1003,14 @@ def time_interval(self) -> biob.types.TimeInterval:
         """
         return self.reader.time_interval
 
+    @property
+    def standard_metadata(self) -> biob.standard_metadata.StandardMetadata:
+        """
+        Return the manifest of standard metadata as a dictionary. The possible
```
SeanLeRoy, line 1009: Not a dictionary so maybe this?
```suggestion
        Return a set of standardized metadata. The possible
```

SeanLeRoy (approved):

bioio/bio_image.py
BrianWhitneyAI: whoops! relic of the past

toloudis (approved):

PR #111, opened 2025-04-03
admin/pin-bioio-base

BrianWhitneyAI:
### Descriptiion 

We have been having issues with releasing new versions of bioio and bioio-base with breaking changes. This allows users to have dependency misalignments. We decided that the simple solution for now is to pin a specific version of bioio-base with every release of bioio. This will ensure that a valid dependency structure is intact.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,8 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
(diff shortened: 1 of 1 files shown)


SeanLeRoy (approved):
Can the comment go above the version rather than same line + below? Besides that LGTM

toloudis (approved):

Issue #104, opened 2025-03-25
Readme still has python 3.9


BrianWhitneyAI:
We dropped support for 3.9 we should remove it from the read me. This may need to happen with readers as well.

Issue #70, opened 2024-08-20
Remove numpy<2.0.0 restriction


pgarrison:
Ideally bioio should support the latest numpy.

I checked with the [ruff rule for numpy 2.0 migration](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#ruff-plugin), which is not (cannot possibly be) 100% comprehensive and found no issues in `bioio`, `bioio-base`, and `bioio-ome-zarr`. Does anyone (@SeanLeRoy) know of anything that would be broken by numpy 2?

If necessary, individual bioio plugins could retain the numpy<2.0.0 restriction until they are updated for compatibility.

BrianWhitneyAI:
Currently fails when unpinning numpy (upgraded to 2.1.1) (Python 3.10) 
```
2024-09-25 09:41:09.904 [error] pytest test discovery error for workspace:  /Users/brian.whitney/Desktop/Repos/bioio 
 ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

Interrupted: 4 errors during collection
 Check Python Test Logs for more details.
2024-09-25 09:41:09.906 [error] pytest test discovery error for workspace:  /Users/brian.whitney/Desktop/Repos/bioio 
 ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
 Check Python Test Logs for more details.

Interrupted: 4 errors during collection
[...]

PR #113, opened 2025-04-04
admin/python-support-3.10+

### LINKED ISSUE #10: timeseries writer tests break in actual use
### Describe the Bug

The first and third tests which are commented out [here](https://github.com/bioio-devs/bioio/pull/9/commits/8552a3c910e2a299f86ebe62e6dd75318532158a#diff-4a26e04cac22023b008db8ba21a4361dbc31343d9e4e708aa09d9818d80cb0cc) (on `test_timeseries_writer `) fail when uncommented because the data array created has inconsistent shapes for the frames output from the `imageio` reader. 

### Expected Behavior

The tests run as expected and the `DefaultReader` is able to read an image created with the parameters of those tests.

### Reproduction

This is reproducible in `aicsimageio` version `4.11.0` if you add
`reader.dask_data.compute()` or `reader.data` to the same test which effectively does what the test in `bioio` is doing (stitches multi-frame images together and stacks them). The resulting error in both cases is the same:
`ValueError: could not broadcast input array from shape (100,100,4) into shape (1,100,100)`
It appears all frames are the same shape in this case **except** the first one.

### Environment

-   OS Version: macOS 12.6.6
-   bioio Version:  developer branch feature/add-writers @ commit hash ______

### LINKED ISSUE #108: Remove python 3.9 from bioio-czi CI & Readme
On further investigation, this might not be needed: bioio dropped python 3.9 support, but bioio-base hasn't, and bioio plugins only depend on bioio-base.

BrianWhitneyAI:
### Description

This PR does a few things all centered around updating our supported python versions to 3.10-3.13. It resolves #108, #104, #70, kind of resolves #10 maybe? And includes the changes from #105 and #80.

1.) Remove all past refrences to 3.9
2.) unpins `imageio[ffmpeg]`, `numpy`, and `tifffile`
3.) Conditionally resolve 'ome_zarr_writer_2.py` dependencies to support all python versions 
4.) define specific dependencies per python version 
5.) update the time series writer tests which were broken and out of date. (this could probably be another PR but was trying to go green)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,9 @@ dependencies = [
   "numpy>=1.21.0",
   "ome-types[lxml]>=0.4.0",
   "ome-zarr>=0.6.1",
-  "ome-zarr-models>=0.1.1",
+  # Use ngff-zarr for Python 3.10 or lower and ome-zarr-models for Python 3.11 or higher.
+  'ngff-zarr>=0.8.2,<0.9.1; python_version < "3.11"',
+  'ome-zarr-models>=0.1.1; python_version >= "3.11"',
   # This version was revoked from PyPi, but in case of using a stale version
   # or a private package index, this will prevent usage
   "scikit-image!=0.23.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,9 +22,10 @@ authors = [
 classifiers = [
   "Development Status :: 4 - Beta",
   "Natural Language :: English",
-  "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
 ]
 dynamic = ["version"]
 dependencies = [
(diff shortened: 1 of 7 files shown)


BrianWhitneyAI:
Moving back to draft to test some dependency issues

BrianWhitneyAI:
I am pretty unsure about my update to the time series writer tests which is broken due to the dependency scramble. The original test is full of TODOs and caveats. I really just wrote a test to correctly assert whats happening currently but I think i am missing some of the complexity of the RGBA portion. Could use some input here.

BrianWhitneyAI:
> Thanks! I should have caught some of these when moved it to 3.9, but others I had no idea existed.
No problem! still not sure its a fix but maybe a step forward! now for the readers...

pyproject.toml
```diff
@@ -33,17 +34,18 @@ dependencies = [
...
   "fsspec>=2022.8.0",
-  "imageio[ffmpeg]>=2.11.0,<2.28.0",
-  "ngff-zarr>=0.8.2,<0.9.1",
-  "numpy>=1.21.0,<2.0.0",
+  "imageio[ffmpeg]>=2.11.0",
+  "numpy>=1.21.0",
   "ome-types[lxml]>=0.4.0",
   "ome-zarr>=0.6.1",
+  # Use ngff-zarr for Python 3.10 or lower and ome-zarr-models for Python 3.11 or higher.
+  'ngff-zarr>=0.8.2,<0.9.1; python_version < "3.11"',
```
SeanLeRoy, line 42: Why less than 0.9.1 still - I thought that was a 3.9 issue?

SeanLeRoy (approved):

pyproject.toml
BrianWhitneyAI: Oh I didn't try it really. the change here was mostly just the 'python_version < "3.11"''

BrianWhitneyAI:
Tests should pass after #116 is merged

PR #118, opened 2025-04-18
Feature #107 Parse file extension that comes before query string

### LINKED ISSUE #107: Support URLs that don't end in the file extension
### Feature Description

Some bioio readers can read images over the internet. I'd like to read images with URLs like this with a query string:
`https://allencell.s3.amazonaws.com/aics/hipsc_12x_overview_image_dataset/stitchedwelloverviewimagepath/05080558_3500003720_10X_20191220_D3.czi?versionId=_KYMRhRvKxnu727ssMD2_fZD5CmQMNw6`

### Use Case

The above example is publicly released AICS data, pinned to a specific version. When referencing data in a versioned AWS S3 bucket, I recommend quantitative analyses pin specific versions of the data for reproducibility.

### Solution

Perhaps bioio could look for a file extension from both `user's path` and `user's path without the query string` 

### Alternatives
Don't rely on file extension and delegate entirely to each reader's `_is_supported_image` which may or may not look at the file extension. This might substantially increase the time to call `BioImage(...)` if each reader's `_is_supported_image` call does I/O.

pgarrison:
# Issue
Some bioio readers can read images over the internet. I'd like to read images with URLs like this with a query string:
`https://allencell.s3.amazonaws.com/aics/hipsc_12x_overview_image_dataset/stitchedwelloverviewimagepath/05080558_3500003720_10X_20191220_D3.czi?versionId=_KYMRhRvKxnu727ssMD2_fZD5CmQMNw6`

The above example is publicly released AICS data, pinned to a specific version. When referencing data in a versioned AWS S3 bucket, I recommend quantitative analyses pin specific versions of the data for reproducibility.

Resolves #107

# Changes
* Now `BioImage.determine_plugin` looks for the file extension at the end of both `user's path` and `user's path without the query string`
* ~This repo (bioio) now has integration tests for bioio+bioio-czi. This is a significant adjustment to the organization of the bioio repository. I think it's right for us because:~
   i. ~we need integration tests at this level since the behavior of `BioImage(path, reader=SomethingSpecific)` and `BioImage(path)` diverge,~
   ii. ~I expect the main code under test by these integration tests will be code from this repo (since tests that don't need bioio will go in individual plugin repos), and~
   iii. ~though bioio-czi integration tests must be licensed under GPLv3, we can license *just* that file under GPLv3 and keep bioio BSD-licensed (bioio-czi + bioio tests depend on bioio-czi, but bioio itself does not).~
* New dummy plugin `accepting-plugin` that pretends to read anything


bioio/tests/test_plugins.py
```diff
@@ -16,7 +16,8 @@ def test_dump_plugins(dummy_plugin: str) -> None:
     old_stdout = sys.stdout
     sys.stdout = StringIO()
     try:
-        dump_plugins()
+        # Disable plugin caching to prevent other tests from interfering
+        dump_plugins(use_cache=False)
```
pgarrison, line 20: This was necessary because the test I added doesn't use the dummy plugin and runs before this test.

kmitcham (approved):

bioio/bio_image.py
```diff
@@ -255,6 +255,28 @@ def determine_plugin(
...
+        >>> BioImage._path_has_extension("./file.ome.zarr", ".zarr")
+        True
+        >>> BioImage._path_has_extension("./file.ome.zarr", ".ome")
+        False
+        """
+
+        def ends_with_ext(string: str) -> bool:
+            return string.lower().endswith(extension)
+
+        return ends_with_ext(path) or ends_with_ext(path.split("?")[0])
```
BrianWhitneyAI, line 278: Is this the only case we are handling? Is this just for S3 files?

pyproject.toml
```diff
@@ -62,6 +62,7 @@ lint = [
   "pre-commit>=2.20.0",
 ]
 test = [
+  "bioio-czi @ git+https://github.com/bioio-devs/bioio-czi.git@main",
```
BrianWhitneyAI, line 65: Is this supposed to point to the repo for integration?

SeanLeRoy (approved):
I think it would make more sense for the readers to have the integration tests (bioio-czi) rather than bioio have integration tests for 13+ readers.

pgarrison:
> I think it would make more sense for the readers to have the integration tests (bioio-czi) rather than bioio have integration tests for 13+ readers. LGTM besides that

If we go that route I'd recommend setting up bioio to run those tests against bioio PRs (on the expectation that those tests are in large part about logic in the bioio code). This is feasible but will require being careful about which branch(es) we test against.

bioio/bio_image.py
pgarrison: This is the only practical use case I know of at this time. It seems plausible that other web servers hosting CZI files might use query strings we'd like to ignore.

pyproject.toml
pgarrison: I imagine if we want to test against a plugin repo, we'd rather test against `main` than the latest release. This gives us more flexibility, like when coordinating ~simultaneous changes in both repos.

pgarrison:
Resolved @SeanLeRoy's feedback by removing the bioio-czi integration test and instead testing the new code with a new dummy plugin. The pre-existing dummy plugin wouldn't work for my test because it raises an exception upon `__init__`.

pgarrison:
I verified that my new test fails without the other changes from this PR.

SeanLeRoy (approved):
✅ nice! quick question: why accepting_plugin underneath accepting-plugin? Is doubly nesting that necessary?

bioio/bio_image.py
BrianWhitneyAI: Nice! this looks good to me then. like the sub functions

bioio/plugins.py
```diff
@@ -265,11 +265,16 @@ def get_plugins(use_cache: bool) -> Dict[str, List[PluginEntry]]:
     return plugins_by_ext
 
 
-def dump_plugins() -> None:
+def dump_plugins(use_cache: bool = True) -> None:
```
BrianWhitneyAI, line 268: Nice!

bioio/tests/accepting-plugin/accepting_plugin/reader.py
```diff
@@ -0,0 +1,40 @@
+from typing import Any, Optional, Tuple
```
BrianWhitneyAI, line 1: Whats the difference here from using the dummy plugin?

BrianWhitneyAI: the plugin name is a little strange to me.

pyproject.toml
BrianWhitneyAI: Ok this works for me then. Initially I would think that we would only want to run it against "approved/passing" releases but I can see the utility here.

BrianWhitneyAI (approved):

BrianWhitneyAI:
> ✅ nice! quick question: why accepting_plugin underneath accepting-plugin? Is doubly nesting that necessary?

probably just so that it can have a init and pyproject / recognize as a package right?

pgarrison:
> ✅ nice! quick question: why accepting_plugin underneath accepting-plugin? Is doubly nesting that necessary?

I think so, since the plugins are installed with [pip install](https://github.com/bioio-devs/bioio/blob/main/bioio/tests/conftest.py#L94). The plugin could be reorganized, e.g. into a single file, but this pattern stays parallel to the existing dummy-plugin and the real plugins.

bioio/tests/accepting-plugin/accepting_plugin/reader.py
pgarrison: The dummy plugin raises an exception when you try to instantiate it (or use it). This one returns. I picked "accepting" because it accepts lots of file types and if you instantiate it with `BioImage(...)` it will accept any arguments. We can certainly rename in follow-up PRs.

PR #119, opened 2025-04-25
feature/v3-ome-zarr-writer

BrianWhitneyAI:
## Initial draft for 0.5.0 Ome Zarr

zarr doesn't support 3.10 anymore so those tests fail.


bioio/writers/generate_zarr.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
```
BrianWhitneyAI, line 1: This file won't go in the repo but is cool to play around with to make your own zarrs

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+        Downsampling factors (t_factor, c_factor ignored, zf, yf, xf).
+
+    Returns
+    -------
+    np.ndarray
+        Downsampled array.
+    """
+    # TODO: There is probably a much more efficent way to do this
+
+    # Only spatial downsampling (time & channel stay intact)
```
SeanLeRoy, line 29: Should we remove the t_factor and c_factor options if we don't use them?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+
+    # Compute how many full blocks fit
+    nz, ny, nx = Z // zf, Y // yf, X // xf
+
+    # Trim off remainders so reshape works
+    d = data[..., : nz * zf, : ny * yf, : nx * xf]
+
+    # Reshape into blocks and average
+    # new shape: (T,C,nz,zf,ny,yf,nx,xf)
+    d = d.reshape(T, C, nz, zf, ny, yf, nx, xf)
```
SeanLeRoy, line 44: hmm wait why do we do this part and the average below?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+
+    Returns
+    -------
+    List[dict]
+        List of axis metadata dictionaries.
+    """
+
+    # TODO: should this be a constant or a util or something?
+    axes = []
+    for n, t, u in zip(names, types, units):
```
SeanLeRoy, line 76: Should we validate these are the same size here?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+            Axis units.
+        axes_scale : List[float]
+            Physical scale per axis.
+        scale_factors : Tuple[int, int, int, int, int]
+            Downsampling factors per axis.
+        num_levels : Optional[int]
+            Maximum number of pyramid levels.
+        chunks : Union[str, Tuple[int, ...], List[Tuple[int, ...]]]
+            Chunk specification or "auto".
+        shards : Optional[Union[Tuple[int, ...], List[Tuple[int, ...]]]]
```
SeanLeRoy, line 133: Can we put the defaults in here? Still kinda feeling like this writer could have a factory method for users that don't have the brain bandwidth or care enough to optimize for the best zarr share for their usage. The simplicity level of at least the .ome.tiff writer is what I'm thinking for the factory method.

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+        channel_names : Optional[List[str]]
+            Channel labels.
+        channel_colors : Optional[List[str]]
+            Channel hex colors.
+        creator_info : Optional[dict]
+            Creator metadata.
         """
-        # Resolve final destination
-        fs, path = biob.io.pathlike_to_fs(uri)
+        self.shape = tuple(shape)
```
SeanLeRoy, line 146: Isn't this already a tuple?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+        self.root = self._init_store(store)
+        self.datasets = self._create_arrays(compressor)
+
+        # Build metadata
+        axes_meta = default_axes(axes_names, axes_types, axes_units)
+        channel_meta = None
+        if channel_names or channel_colors:
+            C = shape[1]
+            channel_meta = []
+            default_colors = [
```
SeanLeRoy, line 167: I think we had different default colors before - where does this default come from?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+            for i in range(C):
+                color = (
+                    channel_colors[i]
+                    if channel_colors and i < len(channel_colors)
+                    else default_colors[i % len(default_colors)]
+                )
+                label = (
+                    channel_names[i]
+                    if channel_names and i < len(channel_names)
+                    else f"Channel {i}"
```
SeanLeRoy, line 184: Is this the OME default naming convention for channel names? I thought it was something like `channel:{index}` but idk could be totally misremembering

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
-                        "min": float(channel_minmax[i][0]),
-                        "start": float(channel_minmax[i][0]),
-                    },
-                }
+        shapes = [self.shape]
+        lvl = 1
+        while max_levels is None or lvl < max_levels:
+            prev = shapes[-1]
+            nxt = tuple(
+                max(1, prev[i] // self.scale_factors[i])
```
SeanLeRoy, line 233: We previously let users choose scale factors for each level (like 0.5, 0.4, 0.3, 0.2) idk how useful that is but is there a reason you removed that?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
-                f"Chunk dimensions:{chunk_dims} do not match data. "
-                f"Expected chunk dimension length:{ndims}"
-            )
+        # TODO: This function kinda sucks, needs better
+        levels = len(self.level_shapes)
 
-        if physical_pixel_sizes is None:
-            pixelsizes = (1.0, 1.0, 1.0)
+        # Auto-chunk only applies when name == "chunks"
+        if param == "auto" and name == "chunks":
```
SeanLeRoy, line 275: Do we benefit from having this be used both for chunks and shards? Feels like it is a bit too generic since we then check if this is for chunks or shards in here

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
-                    "scale": [scale_dim_map[d] for d in dimension_order],
-                }
-            ]
+            items = [param] * levels
+
+        if len(items) != levels:
+            raise ValueError(f"Length of {name} list must be {levels}")
+
+        # Clamp each item to the level shape
+        return [
```
SeanLeRoy, line 296: I'm confused here what `items` is vs `level_shapes`

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
-        plane_size = (
-            image_data.shape[xdimindex]
-            * image_data.shape[ydimindex]
-            * image_data.itemsize
-        )
 
-        target_chunk_size = 16 * (1024 * 1024)  # 16 MB
-        # this is making an assumption of chunking whole XY planes.
+    @staticmethod
+    def _init_store(store: Union[str, zarr.storage.StoreLike]) -> zarr.Group:
```
SeanLeRoy, line 302: If no store is provided can we create one for the user using this?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+        ----------
+        compressor : Optional[BloscCodec]
+            Compressor codec to apply to chunks.
+
+        Returns
+        -------
+        List[zarr.Array]
+            Zarr array objects for each level.
+        """
+        comp = resolver or BloscCodec(
```
SeanLeRoy, line 335: Mentioning this is the default would be good

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+        )
+        arrays: List[zarr.Array] = []
+        for lvl, shape in enumerate(self.level_shapes):
+            chunks_lvl = self.chunks[lvl]
+            shards_lvl = self.shards[lvl]
+            if chunks_lvl is not None and shards_lvl is not None:
+                # safe to zip two tuples
+                shards_param = tuple(c * s for c, s in zip(chunks_lvl, shards_lvl))
+            else:
+                shards_param = None
```
SeanLeRoy, line 346: When would this happen?

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
-                "type": "space",
-                "unit": "micrometer",
-            },
         }
+        ome: Dict[str, Any] = {"multiscales": [multiscale], "version": "0.5"}
+        if channels:
+            ome["omero"] = {"version": "0.5", "channels": channels}
+        if creator:
+            ome["_creator"] = creator
+        self.root.attrs.update({"ome": ome})
```
SeanLeRoy, line 411: Is there a spot in the specification for `plates`? In general I think it would be good if we provide at least enough metadata here to grab out all of the "standard_metadata" fields we like to get. I think just `plates` for the well row/column is missing, but maybe timelapse interval too if the multiscales doesn't track the delta between timepoints.

bioio/writers/ome_zarr_writer.py
```diff
@@ -1,431 +1,478 @@
...
+        Tuple[int, ...]
+            Chunk sizes (t, c, z, y, x).
+        """
+        bpe = self.dtype.itemsize
+        maxe = (64 << 20) // bpe
+        base = int(math.sqrt(maxe))
+        y = min(shape[3], base)
+        x = min(shape[4], maxe // y) or 1
+        z = min(shape[2], max(1, maxe // (y * x)))
+        return (1, 1, z, y, x)
```
SeanLeRoy, line 433: simplify by just making this deal with spatial dims?

bioio/writers/ome_zarr_writer.py
BrianWhitneyAI: yeah this is the same functionality as the V2 writer

bioio/writers/ome_zarr_writer.py
BrianWhitneyAI: zip raises an error if they're not. would validating add some additional functionality here?

bioio/writers/ome_zarr_writer.py
BrianWhitneyAI: Agreed. I think that's another PR though, this ones getting pretty large.

BrianWhitneyAI:
closing this PR in favor of https://github.com/bioio-devs/bioio-ome-zarr/pull/41

PR #126, opened 2025-05-21
feature/migrate-writers

BrianWhitneyAI:
### Link to Relevant Issue

This pull request resolves #


BrianWhitneyAI:
Documentation is a question here. Where to have writer documentation...

bioio/writers/__init__.py
```diff
@@ -1,15 +1,21 @@
...
-]
+
+import sys
+from typing import List
+
+# Choose the right entry_points based on Python version
+if sys.version_info >= (3, 10):
+    from importlib.metadata import entry_points as _entry_points
+else:
+    from importlib_metadata import entry_points as _entry_points
```
kmitcham, line 10: these lines (8 and 10) seem identical?

kmitcham (approved):

bioio/writers/__init__.py
BrianWhitneyAI: resolved in [c4f40b1](https://github.com/bioio-devs/bioio/pull/126/commits/c4f40b166fac3544409d4537935bee62f29cc089)

pyproject.toml
```diff
@@ -31,21 +31,11 @@ dynamic = ["version"]
 dependencies = [
   # We always pin a specific version of bioio-base with
   # each release so that we avoid bioio + bioio-base misalignment.
-  "bioio-base==1.0.7",
+  "bioio-base>=1.0.7",
   "dask[array]>=2021.4.1",
-  "fsspec[http]>=2022.8.0",
```
SeanLeRoy, line 36: I think the dummy plugin reader uses this so you'll want to add it to the optional `test` deps

bioio/bio_image.py
```diff
@@ -1156,7 +1156,17 @@ def save(
         ids in each Image's name attribute but IDs will be generated. The order of the
         scenes will be the same (or whatever order was specified / provided).
         """
-        from .writers import OmeTiffWriter
+        import bioio.writers as _writers  # type: ignore[attr-defined]
+
+        OmeTiffWriter = getattr(_writers, "OmeTiffWriter", None)
+
+        if OmeTiffWriter is None:
```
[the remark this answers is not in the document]
BrianWhitneyAI: We implement a specific save function built into BioImage that defaults to the OmeTiffWriter. It never had the option of the OmeZarrWriter. This just maintains the old functionality but warns the user if OmeTiffWriter isnt installed since it comes from bioio-ome-tiff now.

pyproject.toml
```diff
@@ -31,21 +31,11 @@ dynamic = ["version"]
 dependencies = [
   # We always pin a specific version of bioio-base with
   # each release so that we avoid bioio + bioio-base misalignment.
```
SeanLeRoy, line 33: Update this comment or pin bioio-base? Why the change to pinning it?

bioio/bio_image.py
```diff
@@ -1165,15 +1175,11 @@ def save(
         image_names: List[Optional[str]] = []
         physical_pixel_sizes: List[biob.types.PhysicalPixelSizes] = []
 
-        # Get selected scenes / handle None scenes
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [1f69542](https://github.com/bioio-devs/bioio/pull/126/commits/1f69542024d38c389d1f41a66d09a9ee1d17e0ab)

SeanLeRoy (approved):

bioio/bio_image.py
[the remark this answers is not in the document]
BrianWhitneyAI: yeah this is another argument for a more involved ABC in bioio_base for writers. Then we could implement a way to choose what writer to use for save

pyproject.toml
BrianWhitneyAI: never mind no they didnt. Added back in [38c66f8](https://github.com/bioio-devs/bioio/pull/126/commits/38c66f8fec21c1ae7601058c5ca7bafbd46fb38a)

BrianWhitneyAI:
tests will fail until bioio base 1.0.8 is released and the ABC writer is added.

bioio/bio_image.py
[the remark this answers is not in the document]
toloudis: This was written when OME-TIFF was clearly the only important open file format we wanted, and before we were dealing with multi-TB multiscene time series, nevermind chunked zarr 😄 .    Right now I question whether a single save function in the BioImage class even is useful.  

The code change as implemented here seems to make sense for maintaining existing functionality... 
but I think long term if writers are being removed from bioio, then either (a) `save` should be removed from BioImage, or (b) `save` could get a abstract Writer passed in and the caller would have to do the importing and constructing the writer object.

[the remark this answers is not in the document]
toloudis: Since this moving of writers seems to be a breaking change anyway, maybe this is a good opportunity to remove `save` until we design something better (abstracted api or not)?

bioio/writers/__init__.py
```diff
@@ -1,12 +1,15 @@
...
-__all__ = [
-    "OmeTiffWriter",
-    "OmeZarrWriter",
-    "OmeZarrWriter2",
-]
+# Public API list
+__all__: List[str] = []
+
+# Discover all registered writers
+_eps = _entry_points(group="bioio.writers")
```
toloudis, line 10: how do writers get registered?  This seems to need some documentation somewhere maybe?

BrianWhitneyAI: I'll open another documentation branch once the writers are on main for their respective plugins. That way we can reference directly.

PR #132, opened 2025-06-06
Feature/is supported reporting

kmitcham:
### Link to Relevant Issue

This pull request resolves #
https://github.com/bioio-devs/bioio-bioformats/issues/21

### Description of Changes

Make the exception thrown on unsupported image have useful text

### DIFF
--- a/bioio/array_like_reader.py
+++ b/bioio/array_like_reader.py
@@ -127,8 +127,12 @@ class ArrayLikeReader(Reader):
                 "ArrayLikeReader supported types are numpy ndarray, dask Array,"
                 + "or xarray DataArray."
             )
-
-        return isinstance(image, (np.ndarray, da.Array, xr.DataArray))
+            raise exceptions.UnsupportedFileFormatError(
+                "ArrayLikeReader", ", ".join(errors)
+            )
+        else:
+            return True
+        # If we got here, we have a supported image type
 
     @staticmethod
     def _guess_dim_order(shape: Tuple[int, ...]) -> str:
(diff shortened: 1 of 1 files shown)


bioio/array_like_reader.py
```diff
@@ -110,13 +110,29 @@ class ArrayLikeReader(Reader):
     def _is_supported_image(  # type: ignore
         image: Union[List[MetaArrayLike], MetaArrayLike], *args: Any, **kwargs: Any
     ) -> bool:
+        errors = []
```
BrianWhitneyAI, line 113: A list of strings called `errors` is pretty confusing to me. Why not use real errors?

bioio/array_like_reader.py
```diff
@@ -110,13 +110,29 @@ class ArrayLikeReader(Reader):
...
+        if errors:
+            errors.append(
+                "ArrayLikeReader supported types are numpy ndarray, dask Array,"
+                + "or xarray DataArray."
             )
-
-        return isinstance(image, (np.ndarray, da.Array, xr.DataArray))
+            raise exceptions.UnsupportedFileFormatError(
+                "ArrayLikeReader", ", ".join(errors)
+            )
```
BrianWhitneyAI, line 132: I dont think that we need all of this if else complexity. I think the original `if isinstance` was really just there so we could use `all()`. Could we do something like 
```
SUPPORTED = (np.ndarray, da.Array, xr.DataArray)
imgs = image if isinstance(image, list) else [image]
    for img in imgs:
        if not isinstance(img, SUPPORTED):
            raise exceptions.UnsupportedFileFormatError()
Return True
```

aswallace (approved):

PR #149, opened 2025-08-22
bugfix/reader_dims

### LINKED ISSUE #148: Bug: Injection of dimensions between bioio and readers
Reading an image with BioImage vs the reader directly can produce different dims. 

Image

Underneath BioImage we use the dask data dims and shape instead of the reader dims:

```
    @property
    def dims(self) -> biob.dimensions.Dimensions:
        """
        Returns
        -------
        dims: dimensions.Dimensions
            Object with the paired dimension names and their sizes.
        """
        if self._dims is None:
            self._dims = biob.dimensions.Dimensions(
                dims=self.xarray_dask_data.dims, shape=self.shape
            )

        return self._dims
```

This could also be indicative of an extra padding bug in CZI that doesn't align with its internal defined dims but we should still surface the same dims from reader if possible instead of always formatting xarray_dask_data

BrianWhitneyAI:
## Description 

The purpose of this PR is to resolve #148  and use the specific readers dims instead of building them at the bioio level. reader already inherit dim derivation logic from base.

### DIFF
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -684,12 +684,8 @@ class BioImage(biob.image_container.ImageContainer):
         dims: dimensions.Dimensions
             Object with the paired dimension names and their sizes.
         """
-        if self._dims is None:
-            self._dims = biob.dimensions.Dimensions(
-                dims=self.xarray_dask_data.dims, shape=self.shape
-            )
 
-        return self._dims
+        return self.reader.dims
 
     def get_image_dask_data(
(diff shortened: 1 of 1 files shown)


BrianWhitneyAI:
I think this bug is going to require a policy discussion. Tests are currently failing for dimension prediction mismatches. This is because we pad up to 5D. with this change the inherent datas dims are present without padding 

ex.  for data (1,1)
```
E         - TCZYX
E         + YX
```

Issue #161, opened 2025-12-03
feature-request: Reader priority usage


BrianWhitneyAI:
## Description 

We should allow users to provide a list of existing readers as a order of usage for a file. This should be separate from the `reader=` param. This will allow us to not worry about install order and scenarios where a reader with a large pool of supported images but decreased specialization is used.

PR #162, opened 2025-12-03
feature/plugin-priority

BrianWhitneyAI:
## Description 

The purpose of this PR is to improve our plugin selection. It aims to resolve  https://github.com/bioio-devs/bioio-imageio/issues/29 , https://github.com/bioio-devs/bioio-bioformats/issues/45 and https://github.com/bioio-devs/bioio/issues/161.

We have increasingly become aware that users like to have a lot of plugins installed at the same time. The priority of usage of a readers that support the same file (such as our tiff suites and imageio/bioformats catch-alls) didn't have an easy way for users to control what was used for their files other than specific reader definition. This PR allows for a priority list. 

There was some hidden nuance around the plugin_feasibility_report  (see #) The list of extensions is actually a approved list by bioio, the reader itself can support  more file types (and often will). At runtime BioImage will only read files with the specific instruction. `plugin_feasibility_report` is not bound by extension and will try all readers regardless. This shows things like `bioio-imageio` supporting tiff images. This is intended but not explained. This pr adds a warning to describe the case.

The use case for this is if you have a file with a strange or new ext. you want to see if any of the readers can parse it. you use the `plugin_feasibility_report` to assess the file and see that `bioio-bioformats` can read it! you can then override the extension restriction by passing the the `bioio-bioformats` reader directly.

### Description of Changes 

1. Adds in the `plugin_priority` parameter allowing users to specify a reader priority schema
2. Adds tests to support `plugin_priority`
3. Adds in docs / warning about plugin feasibility to describe "ext not present but supported" file case

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dynamic = ["version"]
 dependencies = [
   # We always pin a specific version of bioio-base with
   # each release so that we avoid bioio + bioio-base misalignment.
-  "bioio-base~=3.1.0",
+  "bioio-base~=3.2.0",
   "dask[array]>=2021.4.1",
   "numpy>=1.21.0",
   "ome-types[lxml]>=0.4.0",


bioio/bio_image.py
```diff
@@ -198,6 +232,7 @@ def determine_plugin(
           extension against the known plugins.
         - For each matching plugin, it tries to instantiate a reader and checks
           if it supports the image.
+        - If multiple plugins match, optional ``plugin_priority`` takes precedence.
```
kmitcham: Is there an explicit definition of what happens with multiple matches, but plugin_priority is not set?  I think we want that, even if the defined order is the first plug alphabetically.

pgarrison: What about this priority order?
* Check extensions in order of length (e.g., if plugins support `.c`, `.b.c`, and `.a.b.c`, then ingesting `file.a.b.c` will first look for plugins supporting `.a.b.c`, then `.b.c`)
  * Among plugins that support the same extension, choose the one that supports the fewest extensions total.

pgarrison (approved):
Nice! I like this approach

SeanDuHare (approved):

bioio/bio_image.py
BrianWhitneyAI: for the above description, the index of the plugin is its priority (0 being highest)

toloudis:
I have a few thoughts I'd like to add.
1. It's good for users to provide a `reader=` arg to BioImage ctor.  This is my favorite way to resolve ambiguities and I would want our documentation to advocate for doing things this way.
1a. In this PR I would advocate to extend the `reader` parameter to allow a list of Reader classes to be passed in optionally, rather than adding a new param for `plugin_priority`. 
2. If the feasibility report behaves differently than what the BioImage constructor does, then we should fix that so that things are consistent/predictable.
3. We need to make it really loud and clear in our docs about the current, existing precedence rules.   We could also give users hints about how to uninstall plugins and reinstall in the order they want.  If it turns out that the date stamp thing is actually unreliable, then we have an actual bug and we really do need to change the default precedence rules.

bioio/bio_image.py
TimMonko: The priority approach of @pgarrison resonates with how I like to think about the various plugins, especially where I want the (presumed) most specific plugin to be prioritized first, and seeing how @BrianWhitneyAI laid it out makes me more certain that its both logical _and_ easily understood with minimal documentation (which, right now is actually quite hard). I've followed y'alls discussions for a while, and while I'm unhappy about the current prioritization (i.e. time-of-install) nothing else has stood out as a compelling alternative until this. 

I think this proposed prioritization logic would cover all edge cases that I keep running in to, but it is nice that there still is this proposed API for a priority override. It will make troubleshooting a set of reader behaviors more quickly than just trying to pass the correct reader function.

toloudis:
@BrianWhitneyAI @pgarrison 

I agree with the basic rules you laid out from Brian's post like this:

1. Fewer extension families first — Plugins that support fewer groups of suffix-related extensions (e.g., .ome.tiff + .tiff is one family) are considered more specific and are ranked higher.
2. If tied, fewer raw extensions first — A plugin advertising fewer total extensions in its get_supported_extensions() list is treated as more focused and ranked higher.
3. If still tied, alphabetical by plugin name — A stable, deterministic tie-breaker.

I guess this requires asking EVERY installed plugin for its list of extensions, and maybe caching it all so you never have to do it a second time.
 
But before all that, I suggest that rule 0 is:
if a single reader is provided in the BioImage constructor, use that reader only.  If a list, then use the list in list order of precedence.  Only if no reader argument is provided, then follow the default precedence rules. 

**Important**: if implemented, then can we DO AWAY WITH INSTALL DATE ORDER? as in get rid of that code for everyone.  Could be a breaking change.
I also stand by making sure that the feasibility report is doing the correct behavior and can emulate the results of the BioImage ctor.

BrianWhitneyAI:
Working on this problem has revealed to me that our current Dummy plugin / Accepting plugin test pattern is pretty inefficient and not very versatile. It also has some build corruption issues that come up within the test multithreading. I am experimenting with alternatives but will open a seperate PR for that.

bioio/plugins.py
```diff
@@ -173,31 +269,101 @@ def get_dependency_version_range_for_distribution(
...
+       be tried *before* a broader, more generic reader.
+
+    Notes
+    -----
+    * Before including a plugin, we check its declared ``bioio-base`` version
+      range against the version required by the core ``bioio`` distribution and
+      skip plugins that are incompatible.
+    * This default ordering controls BioIO's automatic choice of reader based
+      on extension. A user can still override the order of candidate plugins
+      for a given file via the ``plugin_priority`` argument on ``BioImage``.
```
kmitcham: I like these docs.  Complicated, but it is a complicated problem.

BrianWhitneyAI:
I opened https://github.com/bioio-devs/bioio/pull/168 which adds a test suite for this feature. Im pretty happy with how it came out tbh. It drops our real install for synthetic factories and greatly improves test performance. I dont think this PR should be merged without it.

bioio/bio_image.py
```diff
@@ -281,36 +333,69 @@ def ends_with_ext(string: str) -> bool:
...
+
+        * If ``reader`` is ``None``, the default plugin ordering is used.
         """
-        if reader is not None:
-            # Check specific reader image types in a situation where a specified reader
-            # only supports some of the ImageLike types.
-            if not check_type(image, reader):
+
+        # Force specific reader
+        if isinstance(reader, type):
```
toloudis: what is `type` in this statement? I can't see where it comes from.

toloudis: I had to google to figure out what this means in Python but now I understand.

bioio/bio_image.py
```diff
@@ -281,36 +333,69 @@ def ends_with_ext(string: str) -> bool:
...
-                    reader.__name__, str(type(image))
+                    forced_reader.__name__, str(type(image))
                 )
+            return forced_reader(image, fs_kwargs=fs_kwargs, **kwargs), None
 
-            return reader(image, fs_kwargs=fs_kwargs, **kwargs), None
-
-        # Determine reader class based on available plugins
+        # Determine plugin with priority
+        plugin_priority: Optional[Sequence[Type[biob.reader.Reader]]] = None
```
toloudis: maybe you could remove the entire block of code above if you just turn the single reader into a list: `plugin_priority=[reader]` and then fall through to this code?

bioio/plugins.py
```diff
@@ -40,13 +47,73 @@
...
+            e = "." + e
+
+        if e not in seen:
+            seen.add(e)
+            normalized.append(e)
+
+    return normalized
+
+
+def _count_extension_families(exts: Sequence[str]) -> int:
```
toloudis: I feel like unit testing just this function would be very useful; actually, unit testing the results showing in the "Example" below is what I'd be curious about. Testing the counts is a bit simpler, but actually testing that it finds the expected families might be a more interesting test.

bioio/bio_image.py
BrianWhitneyAI: Right now we differentiate the "force plugin" by it essentially not being list type. The Idea being that users can define a single reader and force bioio to try and read the file with that option. If we go to the list type with the current logic than the selected reader would become just top of the priority list and would 1) look at exts and 2) possibly be read by a different reader. I am in favor of leaving the typing as is but am open to other options.

BrianWhitneyAI: IF no reader can read the file then there will be an error. If an installed reader can read it we raise that warning. We could move that warning up to error level?

toloudis: oh I thought that if a user passed a list of readers, then we should only try those readers and then stop.  that seems more predictable from an api standpoint but I can see why you might want to do it the other way too.

bioio/bio_image.py
BrianWhitneyAI: [20cb352](https://github.com/bioio-devs/bioio/pull/162/commits/20cb35270339337ed95750963e5198fba85bd437)

TimMonko (commented):
I wanted time to do a more full review, and I think the biggest opinion I have is in https://github.com/bioio-devs/bioio/pull/162/files#r2595957308 about how to think about the code pathways and trying to avoid confusion.

I did have a go at testing out how this would change things for my design (i.e. my own ordering of plugin priority to overcome the current issues) and really like it! ndev-kit/ndevio#42. I especially like that I can pass `reader=[Reader]` with just one, and it will still use the fallbacks of bioio. This way I can emit a warning to users that their intended reader isn't working, but if they expect certain behavior they can easily "choose" a reader. This is just a better experience for someone working in napari. 

If the proposed deterministic ordering is used (fewer suffixes prioritized as more specific) I will be very happy and overall these changes would be a big improvement to my experience with bioio, and I hope for others.

Thanks for the great work and continued tuning to get, what I think, is a key and fantastic feature. If it goes it as is, that will also work for me :)

bioio/bio_image.py
toloudis: That sounds reasonable to me.  I still would recommend to most users to just use the `reader` argument and guess that it covers most usage.  I agree that it would be some kind of error to specify both things.

Here's a thought: Instead of two separate lists of Readers, in `reader` and `plugin_priority`, instead we accept `reader` and a boolean with a name like `try_other_readers`, which defaults to False ?

1. if `reader` is NOT supplied, then we can ignore `try_other_readers`, and try any and all readers by the new rules.
2. if `reader` is provided, then we try those specified readers first, and check the value of `try_other_readers` to know what to do next.

toloudis: I thought there was some user benefit where automatically trying the others was good.  Like this example:  I have an unknown image, but I also know that I want to specify a precedence just for readers that support TIFF (and test those readers first).  For all other readers, I am willing to let bioio's rules take over.  

On the other hand, if you are willing to have bioio ONLY try the readers specified in the list, I'd be in favor of that. It certainly seems simpler.    So then the logic would be:
1. if `reader` is provided, then try to use that reader, or list of readers, only.  if they don't work, then stop and error or whatever. Failed to open BioImage.
2. if `reader` is not provided, then use our new default plugin precedence.

bioio/bio_image.py
```diff
@@ -35,9 +40,22 @@ class BioImage(biob.image_container.ImageContainer):
...
         A string, Path, fsspec supported URI, or arraylike to read.
-    reader: Optional[Type[Reader]]
-        The Reader class to specifically use for reading the provided image.
-        Default: None (find matching reader)
+    reader: Optional[Union[Type[biob.reader.Reader],
+                           Sequence[Type[biob.reader.Reader]]]]
+        Controls how BioImage selects the underlying Reader:
+
+        * If a **single Reader subclass** is provided (e.g. ``reader=TiffReader``),
+          that reader is used directly and plugin ordering is bypassed.
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [847a9a8](https://github.com/bioio-devs/bioio/pull/162/commits/847a9a8eef5c0f1977bf4318ecd61b37ef1c79d9)

bioio/bio_image.py
TimMonko: If that's what yall agree on, it will work for me. It's always hard to balance giving people more options and keeping a simple, easy to understand API, so I get it and want those decisions to be up to you. 

I think for me the main strength of allowing _both_ a user priority list and the default ordering fallback path is that then downstream libraries can still set preferences for some kind of library code. Say, they normally use `.czi` files, so it makes sense for them to put `bioio-czi` as their reader default. In this scenario though, it could be helpful that there is still the fallback to checking all readers -- in some instance maybe the `bioio-czi` reader is not working for their needs because its a slightly incompatible file, see for eg ndev-kit/ndevio#36 for how there was an ome-validation error. 

In this scenario, we can imagine a non-coding user of this library (whether its a napari plugin or just some script that a colleague sent you) wants to have a solution. For them, maybe installing some different library like `bioio-bioformats` could read this file properly, but the original script/package doesn't assume this is installed, so it wouldn't make sense to pass this reader. If the preference didn't work, then it would make sense to fallback to bioio trying all plugins like normal.

In another scenario, we can imagine that a library/script could pass all known plugins to `reader` in a preferred order, but we can imagine both that a plugin may not be installed, or even that a new plugin like `bioio-czi-fork` is made that solves the problem, but isn't know to that script, so it wouldn't have the opportunity to try and work. 

I suppose if the idea is that people passing a list of preferences are coding themselves, then not doing the fallback makes sense, but as we've seen before with the widely appreciated `napari-aicsimageio`, there are many people just looking for a do-all image reader, but perhaps they would still like to define at least _some_ preferences.

As I said earlier, the current implementation is ok, and there are workarounds, but since the machinery is all there, I'd be in favor of supporting all these code pathway options.

toloudis: @TimMonko thank you so much for all the thoughtful comments here and advocating for what you need.

After a couple discussions among the engineers, I think for now we should go with just the improved default plugin precedence, plus the expanded `reader` argument for the BioImage constructor. 

We can then add the `try_other_readers` option, cleanly, after the fact, if there is high demand for it.

bioio/bio_image.py
```diff
@@ -281,41 +305,94 @@ def ends_with_ext(string: str) -> bool:
     @staticmethod
     def _get_reader(
         image: biob.types.ImageLike,
-        reader: biob.reader.Reader,
+        reader: Optional[
+            Union[Type[biob.reader.Reader], Sequence[Type[biob.reader.Reader]]]
```
toloudis: Maybe at this layer you could simplify by always passing in a sequence of Reader (or `None`)

bioio/tests/test_bio_image.py
```diff
@@ -72,37 +71,189 @@ def test_bioimage_submission_data_reader_type_alignment(
 
 
 def test_bioimage_attempts_s3_read_with_anon_attr(
-    sample_text_file: pathlib.Path, dummy_plugin: str
+    sample_text_file: pathlib.Path,
+    plugin_factory: Callable[[Iterable[TestPluginSpec]], list[EntryPoint]],
```
toloudis: I might be naive about how to write Python tests, but the code looks weird having plugin_factory be an input argument to all the functions.  A type definition to shorten `Callable[[Iterable[TestWriterSpec]], list[EntryPoint]]` would probably be nice for readability.  Maybe name it as `plugin_factory: PluginFactoryFixture` ?

bioio/tests/test_plugins.py
```diff
@@ -1,41 +1,187 @@
...
-    finally:
-        sys.stdout = old_stdout
+    # Capture stdout
+    captured = capsys.readouterr()
+    output = captured.out
 
-    # Check if package name is in the output
-    assert dummy_plugin in output
+    # Assert: plugin name appears in dump
+    assert "dummy_plugin" in output
```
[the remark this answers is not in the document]
BrianWhitneyAI: This test is pre-existing. We could add some specific assertions about the wording but Im not sure what else we would want.

bioio/bio_image.py
```diff
@@ -281,41 +305,94 @@ def ends_with_ext(string: str) -> bool:
...
         )
-        ReaderClass = plugin.metadata.get_reader()
-        return ReaderClass(image, fs_kwargs=fs_kwargs, **kwargs), plugin
 
     def __init__(
         self,
         image: biob.types.ImageLike,
-        reader: Optional[Type[biob.reader.Reader]] = None,
+        reader: Optional[
+            Union[Type[biob.reader.Reader], Sequence[Type[biob.reader.Reader]]]
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [3d3f727](https://github.com/bioio-devs/bioio/pull/162/commits/3d3f727d37d7febaeca58891bacaea6c6024a7d5)

bioio/tests/test_bio_image.py
BrianWhitneyAI: added in [3d3f727](https://github.com/bioio-devs/bioio/pull/162/commits/3d3f727d37d7febaeca58891bacaea6c6024a7d5) Type definition doesn't quite work without some aliasing so I added some protocol support should work the same way

PR #195, opened 2026-06-24
feature/finegrain-region-read

BrianWhitneyAI:
## Description 

This is a subsequent release to the update in bioio-base==3.4.0 with more fine-grained region reading. This PR bumps the version and provides a path to the update

### DIFF
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -904,11 +904,21 @@ class BioImage(biob.image_container.ImageContainer):
 
         See `bioio_base.transforms.reshape_data` for more details.
         """
-        # If no out orientation, simply return current data as dask array
+
+        # direct reader sub-region read
+        if (
+            dimension_order_out is not None
+            and kwargs
+            and not (
+                self._reconstruct_mosaic
+                and biob.dimensions.DimensionNames.MosaicTile in self.reader.dims.order
+            )
+        ):
+            return self.reader.get_image_data(dimension_order_out, **kwargs)
+
+        # Otherwise read the full image: return it as-is, or reshape/slice it.
         if dimension_order_out is None and not kwargs:
             return self.data
-
-        # Transform and return
         return biob.transforms.reshape_data(
             data=self.data,
             given_dims=self.dims.order,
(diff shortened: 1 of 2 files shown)


bioio/bio_image.py
```diff
@@ -904,11 +904,21 @@ def get_image_data(
 
         See `bioio_base.transforms.reshape_data` for more details.
         """
-        # If no out orientation, simply return current data as dask array
+
+        # direct reader sub-region read
+        if (
+            dimension_order_out is not None
+            and kwargs
```
kmitcham: so the presence of any extra kwargs could trigger this?  I assume we want some specific ones in this case.

kmitcham (approved):

bioio/bio_image.py
BrianWhitneyAI: This is true will fix!

bioio/bio_image.py
```diff
@@ -904,11 +904,21 @@ def get_image_data(
 
         See `bioio_base.transforms.reshape_data` for more details.
         """
-        # If no out orientation, simply return current data as dask array
+
+        # direct reader sub-region read
```
SeanDuHare: Can you expand this comment - these boolean checks are a bit wonky to read

SeanDuHare (approved):

bioio/bio_image.py
BrianWhitneyAI: yeah will do. Essentially we are inverting the lower checks to say, if its a slice and not weird dims like mosaic then use the direct path

bioio/bio_image.py
BrianWhitneyAI: Actually the only accepted kwargs are slice per the function definition so I think its fine to assert directly here. I added a comment about it though

bioio/bio_image.py
BrianWhitneyAI: split up and updated in [95f5bb4](https://github.com/bioio-devs/bioio/pull/195/commits/95f5bb4b889a7be10963fca9954ea6478b13cfad)

Issue #198, opened 2026-06-24
Broaden use of the reader slice/region fast path across reshape and xarray_dask_data call sites


BrianWhitneyAI:
## Summary

`BioImage.get_image_data` recently gained a slice fast path that defers a sub-region read directly to `reader.get_image_data(...)` instead of materializing the full array and slicing in memory (see `bioio/bio_image.py` ~L913–916, from #195). Right now that fast path only fires under a narrow condition. There are several other places in the repo where we either read the full image and then reshape/slice, or interact with `xarray_dask_data` directly, that could instead fall back to the reader — which often has a faster, format-native way to derive the result.

This issue tracks broadening use of the reader slice/region fast path (and reader fallbacks generally) across these call sites.

## Existing fast path

```python
# bioio/bio_image.py — get_image_data
if dimension_order_out is not None and kwargs and not stitching_mosaic:
    return self.reader.get_image_data(dimension_order_out, **kwargs)

# Otherwise read the full image: return it as-is, or reshape/slice it.
if dimension_order_out is None and not kwargs:
    return self.data
return biob.transforms.reshape_data(
    data=self.data,            # <-- full in-memory read, then slice
    given_dims=self.dims.order,
    return_dims=dimension_order_out,
    **kwargs,
)
```

## Target areas

1. **`get_image_data` with slice kwargs but no `dimension_order_out`** (`bio_image.py` ~L915–926). When a caller passes only slice kwargs (e.g. `get_image_data(C=0)`), the fast-path condition (`dimension_order_out is not None`) is not met, so we fall through to a full `self.data` read + `reshape_data`. We could default `dimension_order_out` to the natural order and still route through the reader slice fast path.

2. **`get_image_dask_data`** (`bio_image.py` ~L834). Mirrors `get_image_data` but builds from `xarray_dask_data`. Evaluate an analogous reader fallback (`reader.get_image_dask_data`) so delayed slicing can use a reader-native region path where available.

3. **`reshape_data` on full `arr.data`** in `_transform_data_array_to_bioio_image_standard` (`bio_image.py` ~L519–524) and the `get_image_data` fallback (~L921). Where the reshape is purely a slice/reorder, prefer pushing the selection down to the reader rather than reshaping a fully-materialized array.

4. **`channel_names`** (`bio_image.py` ~L1056). Currently pulls `self.xarray_dask_data[Channel].values`, which forces task-graph construction just to read channel coordinate labels. Add a reader fallback that derives channel names from metadata.

[...]

Issue #197, opened 2026-06-24
Add fast-path access for metadata properties (dims, shape, dtype) without building full dask task graph


BrianWhitneyAI:
## Summary

Properties derived from `xarray_dask_data` — namely `dims`, `shape`, and `dtype` — currently force construction of the full delayed dask task graph just to read lightweight metadata. Building that task graph can be computationally expensive, yet several readers can access these values cheaply from their underlying file metadata. We should add a fast path so requesting metadata-only properties doesn't pay the cost of assembling the delayed array.

## Current behavior

In both `BioImage` (`bioio/bio_image.py`) and the base `Reader` (`bioio_base/reader.py`), these properties route through `xarray_dask_data`:

- `dtype` → `self.xarray_dask_data.dtype`
- `shape` → `self.xarray_dask_data.shape`
- `dims` → `Dimensions(dims=self.xarray_dask_data.dims, shape=self.shape)`

The first access to `xarray_dask_data` triggers full task-graph construction (and mosaic stitching logic), which is overkill when the caller only wants `shape`/`dtype`/`dims`.

## Proposed change

Introduce an optional fast path for metadata access:
- Allow readers to override lightweight `dims`/`shape`/`dtype` (and ordering) directly from their native metadata, falling back to the existing `xarray_dask_data`-derived implementation when not provided.
- `BioImage` should prefer the reader's fast path when available.

## Relevant readers with a known fast path

- `bioio-czi` (CZI)
- `bioio-ome-zarr` (OME-Zarr)
- `bioio-nd2` (ND2)

These formats expose dimension sizes / dtype in metadata without needing the full delayed array.

## Notes / considerations

- Keep the default behavior unchanged for readers that don't implement the fast path.
- Ensure correctness for mosaic-stitched images, where `BioImage.shape`/`dims` reflect the stitched result rather than raw reader dims.
- Coordinate the base-class interface change in `bioio-base` with the plugin readers above.

## Tracking checklist

- [ ] `bioio-base`: add optional fast-path interface — bioio-devs/bioio-base#72
- [ ] `bioio`: prefer reader fast path in `BioImage` (this repo)
- [ ] `bioio-czi`: implement fast path — bioio-devs/bioio-czi#104
- [ ] `bioio-ome-zarr`: implement fast path — bioio-devs/bioio-ome-zarr#149
- [ ] `bioio-nd2`: implement fast path — bioio-devs/bioio-nd2#47

PR #218, opened 2026-08-19
Prefer reader fast paths for metadata and sub-region reads

BrianWhitneyAI:
## Summary

Implements the `bioio`-side work for #197 and broadens the reader slice/region fast path from #198.

**Metadata fast path (#197)**
- `dims` / `shape` / `dtype` are derived  from `reader` without constructing the standardized delayed array
- `channel_names` prefers `reader.channel_names`, generating the default OME channel id when the reader has no channel dim

**Broadened slice fast path (#198)**
- `get_image_data` and `get_image_dask_data` now defer to the reader for selection-only calls (uses _read_indexed)

## Verification

- Correctness checked against real files in `bioio-czi` / `bioio-nd2` test resources (9 file/config combos including stitched mosaics via both the pylibczirw and aicspylibczi paths, and `reconstruct_mosaic=False`): fast-path metadata matches task-graph-derived values, and all read patterns are byte-identical to slices of the fully materialized (stitched) data.
- Also verified on a 39 GB production ND2 (T=96 C=2 Z=103 Y=1024 X=1024) over a network mount.

## Benchmarks (vs `main`, same plugin versions)

39 GB ND2 over network mount:

| Operation | main | this PR |
|---|---|---|
| `dims`/`shape`/`dtype` | 1.36 s | 74 ms |
| dask plane read + compute | ~1.5 s | 35 ms |
| `get_image_data("YX", T,C,Z)` | 0.36 s | 41 ms |
| `get_image_data("YX")` (order-only) | 10min + (failed locally OOM) | 40 ms |

Local CZI/ND2 test files show the same shape of wins (e.g. order-only plane read on a 165 MB CZI: 195 ms → 7 ms). `channel_names` timing is unchanged until the plugins override it (their #197 checklist item).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved image metadata access without loading the full image into memory.
  * Optimized subregion reads for compatible NumPy and Dask workflows.
  * Reduced unnecessary reshaping and data processing during image access.

* **Reliability**
  * Added stronger validation for image dimensions and subregion specifications.
  * Improved fallback handling when channel metadata is unavailable.

### DIFF
--- a/bioio/bio_image.py
+++ b/bioio/bio_image.py
@@ -497,24 +497,29 @@ class BioImage(biob.image_container.ImageContainer):
         self._xarray_data = None
         self._dims = None
 
+    @staticmethod
+    def _select_standard_dim_order(dims: Sequence[str]) -> str:
+        """
+        Select the BioImage dimension order to standardize the
+        provided dims into
+        """
+        if (
+            biob.dimensions.DimensionNames.Samples in dims
+            and biob.dimensions.DimensionNames.MosaicTile in dims
+        ):
+            return biob.dimensions.DEFAULT_DIMENSION_ORDER_WITH_MOSAIC_TILES_AND_SAMPLES
+        if biob.dimensions.DimensionNames.Samples in dims:
+            return biob.dimensions.DEFAULT_DIMENSION_ORDER_WITH_SAMPLES
+        if biob.dimensions.DimensionNames.MosaicTile in dims:
+            return biob.dimensions.DEFAULT_DIMENSION_ORDER_WITH_MOSAIC_TILES
+        return biob.dimensions.DEFAULT_DIMENSION_ORDER
+
     def _transform_data_array_to_bioio_image_standard(
         self,
         arr: xr.DataArray,
     ) -> xr.DataArray:
         # Determine if we need to add optionally-standard dims
-        if (
-            biob.dimensions.DimensionNames.Samples in arr.dims
-            and biob.dimensions.DimensionNames.MosaicTile in arr.dims
-        ):
-            return_dims = (
-                biob.dimensions.DEFAULT_DIMENSION_ORDER_WITH_MOSAIC_TILES_AND_SAMPLES
-            )
-        elif biob.dimensions.DimensionNames.Samples in arr.dims:
-            return_dims = biob.dimensions.DEFAULT_DIMENSION_ORDER_WITH_SAMPLES
-        elif biob.dimensions.DimensionNames.MosaicTile in arr.dims:
-            return_dims = biob.dimensions.DEFAULT_DIMENSION_ORDER_WITH_MOSAIC_TILES
-        else:
-            return_dims = biob.dimensions.DEFAULT_DIMENSION_ORDER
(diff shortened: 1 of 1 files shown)


kmitcham (approved):

bioio/bio_image.py
```diff
@@ -603,6 +608,18 @@ def resolution_level_dims(self) -> Dict[int, Tuple[int, ...]]:
         """
         return self.reader.resolution_level_dims
 
+    @property
+    def _stitching_mosaic(self) -> bool:
```
SeanDuHare, line 612: can this become boolean oriented? more of a personal preference ig, but IMO it is more readable when something like `_is_stitching_mosaic`

bioio/bio_image.py
```diff
@@ -734,7 +741,10 @@ def dtype(self) -> np.dtype:
         dtype: np.dtype
             Data-type of the image array's elements.
         """
-        return self.xarray_dask_data.dtype
+        if self._xarray_dask_data is not None:
+            return self._xarray_dask_data.dtype
```
SeanDuHare, line 745: Is this ever different than `self.reader.dtype`? Can we just return `self.reader.dtype` always?

bioio/bio_image.py
```diff
@@ -755,9 +765,23 @@ def dims(self) -> biob.dimensions.Dimensions:
             Object with the paired dimension names and their sizes.
         """
         if self._dims is None:
-            self._dims = biob.dimensions.Dimensions(
-                dims=self.xarray_dask_data.dims, shape=self.shape
-            )
+            if self._xarray_dask_data is None and not self._stitching_mosaic:
```
SeanDuHare, line 768: Slightly worried about the idea that the dims could differ before and after the dask array is initialized - how confident are we that Mosaic is the only potential case for this?

SeanDuHare (commented):
I like the spirit of this a lot - I'm worried about the difference in output some of these have the potential for maintenance/debugging wise. See comment for a specific instance of this

bioio/bio_image.py
toloudis: I like this too.. adding "is_" helps with little extra bit of readability

bioio/bio_image.py
```diff
@@ -744,7 +754,7 @@ def shape(self) -> Tuple[int, ...]:
         shape: Tuple[int, ...]
             Tuple of the image array's dimensions.
         """
-        return self.xarray_dask_data.shape
+        return self.dims.shape
```
toloudis: Why do you check for the xarray_dask_data in the dtype function but not here?

bioio/bio_image.py
BrianWhitneyAI: yeah we can, will fix

bioio/bio_image.py
BrianWhitneyAI: I cant think of anything else that would be different, there could be other dims that are associated with mosiac but itd still work the same way

SeanDuHare (approved):

